PR #2, opened 2017-04-10
10.0 project invoicing

ghost:


project_invoicing/models/account_analytic_line.py
```diff
@@ -16,36 +16,57 @@ class AccountAnalyticLine(models.Model):
...
-            else:
-                line.partner_invoice_id = None
+    @api.model
+    def create(self, vals):
+        line = super(AccountAnalyticLine, self).create(vals)
+        if line.project_id:
+            line.partner_invoice_id = line.project_id.partner_id
+        return line
+
+    @api.model
```
joaoalf: @dufresnedavid I think that the decorator here is `api.multi`

project_invoicing/models/project_project.py
```diff
@@ -19,12 +19,22 @@ class ProjectProject(models.Model):
     def generate_invoices(self, data):
         self.ensure_one()
 
-        lines_by_partner = defaultdict(list)
+        self = self.with_context(
+            force_company=self.company_id.id,
```
joaoalf, line 23: @dufresnedavid Why `company_id` and `force_company_id`?

joaoalf (changes requested):

project_invoicing/models/project_project.py
ghost: to ensure that the correct values are taken for multi-company properties used (on partners and products) when generating the invoice. This ensures that a user with multi company access may trigger an invoice from a project related to another company.

project_invoicing/models/account_analytic_line.py
ghost: @joaoalf right

project_invoicing/models/project_project.py
ghost: Also, force_company and company_id are both used by odoo at different places.

joaoalf (approved):

PR #10, opened 2017-04-24
Fix project_stage_no_quick_create

ghost:


ventilooo (changes requested):
why remove `project_stage_no_quick_create/views/project_project.xml` ?

ventilooo (approved):

PR #34, opened 2017-12-04
Traceback création tâche Dispatch

ventilooo:
* Add `False` as default value for user_id
* Set html field content to `utf-8`

### DIFF
--- a/project_task_autocreate/models/project_task_template.py
+++ b/project_task_autocreate/models/project_task_template.py
@@ -286,7 +286,7 @@ class ProjectTaskTemplate(models.Model):
         """
         description = self.description
         if description:
-            description = description.format(object=record)
+            description = unicode(description).format(object=record)
         vals = {
             'name': self.name.format(object=record),
             'description': description,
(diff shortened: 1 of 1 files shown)


project_task_autocreate/models/project_task_template.py
```diff
@@ -270,7 +270,7 @@ def get_task_vals(self, record=None):
             'priority': self.priority,
             'tag_ids': [(6, 0, self.tag_ids.ids)],
             'date_deadline': self._get_deadline(record),
-            'user_id': self._get_user_id(record).id,
+            'user_id': self._get_user_id(record),
```
joaoalf: @ventilooo If you look carefully at the `get_user_id`, `get_partner_id` and `get_project_id` you will see that they all return a record. Keep that consistency.

ghost:
@ventilooo this can be simplified. Also, you should also fix _get_partner_id and _get_project_id.

rimbendhaou (commented):
@ventilooo If description contains empty fields values it displays False. Is it accepted? It is better to not display False in the discription (just put empty values)

rimbendhaou:
@ventilooo Is it possible to have validation of the description to check its value before saving the task template?

ventilooo:
@joaoalf I'd like your opinion on @rimbendhaou feedbacks.
* Since `description` is an HTML field, how hard would it be to add validation ? and what should be considered valid ?
* For the `False` , I suppose that once the field will be filled with a correct value the false displayed in the task name/description will be overwrite?

ghost (approved):

PR #37, opened 2018-04-11
[10.0]Fix empty team and project

rimbendhaou:
Error message displayed if team is empty or project is empty in the model of the template task

### DIFF
--- a/project_task_autocreate/models/project_task_template.py
+++ b/project_task_autocreate/models/project_task_template.py
@@ -286,7 +286,7 @@ class ProjectTaskTemplate(models.Model):
         """
         description = self.description
         if description:
-            description = str(description).format(object=record)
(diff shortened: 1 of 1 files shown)


project_task_autocreate/models/project_task_template.py
```diff
@@ -286,7 +286,7 @@ def get_task_vals(self, record=None):
         """
         description = self.description
         if description:
-            description = unicode(description).format(object=record)
+            description = str(description).format(object=record)
```
joaoalf: @rimbendhaou The `unicode()` was added to solve a encoding issue when using accents. If you remove it it will cause a regression.

joaoalf (changes requested):

joaoalf (approved):
lgtm

PR #45, opened 2018-08-27
[FIX] Fix database access issue (use dockerize)

joaoalf:
- Use out own customized image (odoo-ci)
- Use dockerize to check if database is alive


rimbendhaou:
Looks ok

rimbendhaou (approved):
ok

PR #47, opened 2018-08-27
[FIX] Fix the task and activity vreation bug

rimbendhaou:
Fix the task and activity vreation bug


joaoalf (approved):
LGTM

you can ignore the codacy warning.

PR #58, opened 2018-09-07
[FIX]Tree view order by starting date attribute

jananjoy:
Ordering the activity record order by "starting date"

### DIFF
--- a/project_event/models/project_task.py
+++ b/project_event/models/project_task.py
@@ -8,6 +8,7 @@ from odoo import api, fields, models
 class Task(models.Model):
     _name = "project.task"
     _inherit = ['project.task']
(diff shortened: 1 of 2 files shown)


project_event/models/project_task.py
```diff
@@ -8,6 +8,7 @@
 class Task(models.Model):
     _name = "project.task"
     _inherit = ['project.task']
+    _order = 'date_start asc'
```
rimbendhaou, line 11: @jananjoy we dont need to order all the tasks and activities by date_start on all the tree views

rimbendhaou:
we don't need to order all the tasks and activities by date_start on all the tree views

PR #59, opened 2018-09-07
[FIX] Remove default 'create and edit'  option from  activity creation form

jananjoy:


rimbendhaou:
@jananjoy In github we must write in english

rimbendhaou (approved):
Lgtm

PR #64, opened 2018-09-20
[ADD] New fields in the Task Model

jananjoy:


eilst:
Implemnted as room_id and status

PR #67, opened 2018-09-21
[ADD] Add resource and room to activity and task template

rimbendhaou:
Add room and resources to task and activity template

### DIFF
--- a/project_event/models/activity_template.py
+++ b/project_event/models/activity_template.py
@@ -2,7 +2,7 @@
 # © 2018 Savoir-faire Linux
 # License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
 
-from odoo import fields, models
+from odoo import api, fields, models
 
 
 class ActivityTemplate(models.Model):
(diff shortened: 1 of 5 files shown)


project_event/models/activity_template.py
```diff
@@ -26,9 +28,40 @@ class ActivityTemplate(models.Model):
...
+    def action_initialize(self):
+        self.action_clear()
+        task = {
+            'room_id': self.room_id.id,
+            'resource_type': 'room',
+            'name': self.name,
+            'activity_template_id': self.id,
+        }
+        self.env['task.template'].create(task)
+        for instrument in self.room_id.instruments_ids:
```
eilst: Exception for: no instruments in a room

eilst: So there is no ids...this will throw an error

rimbendhaou: @eilst if there is no instrument_ids it will not enter in the for clause

eilst (approved):

PR #70, opened 2018-09-26
[FIX] Project Event code enhancement

rimbendhaou:
Project Event code enhancement


project_event/models/project_task.py
```diff
@@ -171,6 +157,16 @@ def action_request(self):
 
     @api.multi
     def action_option(self):
+        calendar_event = self.env['calendar.event']
```
eilst, line 170: This should be in action_request

eilst (approved):

PR #79, opened 2018-10-01
[FIX] Add the French translation version of the list of categories

jananjoy:
Adding the attribute translate into the task.category model


rimbendhaou (approved):
Looks ok

PR #90, opened 2018-10-03
[ADD] Make the fields mandatory in task form

jananjoy:
Make the fields mandatory in the task form

### DIFF
--- a/project_event/views/project_task_view.xml
+++ b/project_event/views/project_task_view.xml
@@ -98,14 +98,14 @@
                             <field name="activity_task_type" invisible="1"/>
                         </group>
                         <group>
-                            <field name="date_start"/>
-                            <field name="date_end"/>
+                            <field name="date_start" required="1"/>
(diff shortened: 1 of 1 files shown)


project_event/views/project_task_view.xml
```diff
@@ -98,14 +98,14 @@
                             <field name="activity_task_type" invisible="1"/>
                         </group>
                         <group>
-                            <field name="date_start"/>
-                            <field name="date_end"/>
```
eilst, line 102: Better go in both server side constraint and here, required

eilst (approved):

project_event/views/project_task_view.xml
jananjoy: @eilst: Dual validation will add performance issues. However, for this requirement, client-side validation is more than enough. :+1: [I like the comment]

rimbendhaou (approved):

PR #98, opened 2018-10-05
[FIX] Corrections to the existing chatter of the event

jananjoy:
Corrections to the existing chatter of the event
Responsible field is added in the chatter with the correct field value 
Title field changed  with correct label


rimbendhaou (approved):
looks good

PR #99, opened 2018-10-05
[ADD] Notes to EV,AC,TK

eilst:


project_event/views/project_project_view.xml
```diff
@@ -55,7 +55,10 @@
                             'default_project_id': active_id,
                         }" domain="[('activity_task_type','=', 'activity')]"/>
                         </page>
-                        <page string="Description">
+                          <page string="Description">
```
rimbendhaou, line 58: @eilst indentation

rimbendhaou (changes requested):
@eilst add translation also

PR #101, opened 2018-10-09
[FIX] Fix wizard event creation from template

rimbendhaou:
Fix wizard event creation from template

### DIFF
--- a/project_event/wizard/project_activity_wizard.py
+++ b/project_event/wizard/project_activity_wizard.py
@@ -29,9 +29,16 @@ class ProjectActivityWizard(models.TransientModel):
         string='Category',
     )
     room_id = fields.Many2one(
-        'resource.resource',
+        'resource.calendar.room',
         string='Room',
     )
-    notes = fields.Text(
+    date_start = fields.Datetime(
+        string='Starting Date',
+        default=fields.Datetime.now,
+    )
+    date_end = fields.Datetime(
+        string='Ending Date',
+    )
+    notes = fields.Html(
         string='Notes',
     )
--- a/project_event/wizard/project_event_wizard.py
+++ b/project_event/wizard/project_event_wizard.py
@@ -2,6 +2,7 @@
 # License LGPL-3.0 or Later (http://www.gnu.org/licenses/lgpl).
 
 from odoo import api, fields, models, _
+from dateutil.relativedelta import relativedelta
 
 
 class ProjectEventWizard(models.TransientModel):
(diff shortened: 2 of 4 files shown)


project_event/wizard/project_task_wizard.py
```diff
@@ -52,10 +52,10 @@ class ProjectTaskWizard(models.TransientModel):
         'hr.department',
         string='Department',
     )
-    duration = fields.Float(
+    duration = fields.Integer(
```
eilst, line 55: 1.5 hours for example, doesn't work?

project_event/wizard/project_event_wizard.py
```diff
@@ -20,6 +21,10 @@ class ProjectEventWizard(models.TransientModel):
         'res.partner',
         string='Responsible',
     )
+    event_partner_id = fields.Many2one(
+        'res.partner',
+        string='Client',
+    )
     event_notes = fields.Text(
         string='Notes',
     )
```
eilst, line 30: In the method add_task():

```
 for activity in activities:
    map(lambda self, task: env['project.task.wizard'].create({
'event_wizard_id': self.id,
                     'template_id': task.id,	                   
                     'activity_wiz_id': act.id,	                   
                     'task_name': task.name,	                     
                     'task_resp_id': task.temp_resp_id.id,	                     
                     'task_category_id': task.task_category_id.id,	                    
                     'resource_type': task.resource_type,	                     
                     'equipment_id': task.equipment_id.id,	                    
                     'room_id': task.room_id.id,	                    
                     'department_id': task.department_id.id,	                    
                     'duration': task.duration,	                   
                     'start_time': task.start_time,	                     
                     'notes': task.notes,	                    
}), activity.template_id.task_template_ids)

```



the nested for loop could be better written like `this:`
map is useful when you want to apply the function to every item of an iterable and return a list of the results. This is simpler and more concise than using a for loop

project_event/wizard/project_activity_wizard.py
```diff
@@ -29,9 +29,16 @@ class ProjectActivityWizard(models.TransientModel):
...
     )
+    date_start = fields.Datetime(
+        string='Starting Date',
+        default=fields.Datetime.now,
+    )
+    date_end = fields.Datetime(
+        string='Ending Date',
+    )
     notes = fields.Text(
         string='Notes',
```
eilst, line 43: I think this would be better:
notes = fields.Html(string='Notes')

project_event/wizard/project_task_wizard.py
rimbendhaou: @eilst it is in minutes only, but it would be better to add a this info in the help or after the field to mention that the unit is in minutes

eilst (approved):

PR #103, opened 2018-10-10
[ADD] Add task creation

eilst:


project_event/models/project_task.py
```diff
@@ -122,9 +122,22 @@ def create(self, vals):
             if vals['activity_task_type'] == 'activity':
```
rimbendhaou, line 129: @eilst else? if activity_task_type not in val? there is no return_create !

project_event/models/project_task.py
```diff
@@ -122,9 +122,22 @@ def create(self, vals):
...
                 vals['code'] = self.env['ir.sequence'] \
                     .next_by_code('project.task.activity')
+                return_create = super(Task, self).create(vals)
+                vals['parent_id'] = return_create.id
+                vals['message_follower_ids'] = None
```
rimbendhaou, line 134: @eilst the is an error when we create event,activity,task from template(with wizard).
ERROR: duplicate key value violates unique constraint "mail_followers_mail_followers_res_partner_res_model_id_uniq"

project_event/models/project_task.py
```diff
@@ -122,9 +122,22 @@ def create(self, vals):
...
+                vals['project_id'] = None
+                self.create_main_task(vals)
             elif vals['activity_task_type'] == 'task':
                 vals['code'] = self.env['ir.sequence'] \
                     .next_by_code('project.task.task')
+                return_create = super(Task, self).create(vals)
+        return return_create
+
+    @api.multi
+    def create_main_task(self, vals):
```
rimbendhaou, line 146: @eilst there is already a main task created from template, main task is duplicated in this case

rimbendhaou (changes requested):

project_event/models/project_task.py
eilst: But what about if there is no template

project_event/models/project_task.py
eilst: Well, it is always there because it is either a task or an activity, it cannot be something else. isn't?

project_event/models/project_task.py
eilst: Done

project_event/models/project_task.py
eilst: Done

project_event/models/project_task.py
rimbendhaou: @eilst we have to check if we are creating from template or not

rimbendhaou: @eilst we have to check if we are creating from template or no

project_event/models/project_task.py
rimbendhaou: @eilst this prob was not resolved

project_event/models/project_task.py
eilst: done

project_event/models/project_task.py
eilst: Done

rimbendhaou (approved):
looks ok

PR #117, opened 2018-10-15
 [ADD] Option state buttons and messages

eilst:


project_event/models/project_task.py
```diff
@@ -239,3 +241,15 @@ def action_read(self):
     def action_draft(self):
         self.write({'task_state': 'draft'})
         self.draft_resources_reservation()
+
+    def send_message(self, action):
+        switcher = {
+            "draft": "",
+            "option": "The following are Optional and no longer on your calendars",
```
rimbendhaou: @eilst we will allow double reservation?

rimbendhaou: @eilst The following are or is?

project_event/models/project_task.py
```diff
@@ -239,3 +241,15 @@ def action_read(self):
...
+            "requested": "The following is requested",
+            "accepted": "",
+            "read": "",
+            "done": "",
+            "canceled": "The following is canceled and no longer on your calendars"
+        }
+        switcher.get(action)
```
rimbendhaou: @eilst when it will send the message?

rimbendhaou (changes requested):

project_event/models/project_task.py
```diff
@@ -162,6 +162,8 @@ def action_request(self):
     def action_option(self):
         if self.activity_task_type == 'task':
             self.draft_resources_reservation()
+        if self.task_state in ['requested', 'read', 'accepted']:
```
rimbendhaou: @eilst I think it should be under the if self.activity_task_type == 'task' statement

project_event/models/project_task.py
```diff
@@ -239,3 +241,32 @@ def action_read(self):
     def action_draft(self):
         self.write({'task_state': 'draft'})
         self.draft_resources_reservation()
+
+    def get_message_body(self, action):
+        switcher = {
+            "draft": "",
+            "option": "The following are Optional and no longer on your calendars",
```
rimbendhaou: @eilst too much long line (PEP8). Add the _('') for translation

project_event/models/project_task.py
```diff
@@ -239,3 +241,32 @@ def action_read(self):
...
+            "read": "",
+            "done": "",
+            "canceled": "The following is canceled and no longer on your calendars"
+        }
+        return switcher.get(action)
+
+    def get_message(self, action):
+        return {
+            'body': self.get_message_body(action),
+            'channel_ids': [(6, 0, [5])],
```
rimbendhaou: @eilst put the external id of the channel not the id

rimbendhaou (changes requested):

project_event/models/project_task.py
eilst: right

eilst: right

rimbendhaou (approved):

PR #125, opened 2018-10-22
[ADD] States Complete

eilst:


project_event/models/project_project.py
```diff
@@ -57,20 +57,32 @@ def _compute_event_log_count(self):
 
     @api.multi
     def action_cancel(self):
+        if self.task_state == 'accepted':
```
rimbendhaou, line 60: there is no cancel of the calendar.event to cancel resource reservation?

project_event/models/project_project.py
```diff
@@ -57,20 +57,32 @@ def _compute_event_log_count(self):
...
         self.write({'state': 'option'})
 
     @api.multi
     def action_draft(self):
         self.write({'state': 'draft'})
 
+    @api.multi
+    def action_postpone(self):
+        if self.task_state == 'accepted':
+            self.send_message('postponed')
```
rimbendhaou, line 89: what about resource reservation?

project_event/models/project_project.py
rimbendhaou: if an event is canceled its activities and tasks must also be canceled.
Same thing for activity (if it is canceled, its tasks should be canceled)

rimbendhaou (changes requested):

project_event/models/project_project.py
eilst: Yes I m adding this functionality in a moment

eilst: Done

project_event/models/project_project.py
eilst: What?

eilst: Ok, I get it. Done

rimbendhaou (approved):

PR #128, opened 2018-10-23
[DEV] Changing the List view of the Task model (#261)

jananjoy:
Ajout de la colonne '*' dans la vue LISTE de la TÂCHE

### DIFF
--- a/project_event/views/project_task_view.xml
+++ b/project_event/views/project_task_view.xml
@@ -11,8 +11,9 @@
                             type="object"/>
                     <button name="action_accept" string="Accept"
                             attrs="{'invisible': [('task_state', '=', 'accepted')]}" type="object"/>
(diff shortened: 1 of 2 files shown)


project_event/views/project_task_view.xml
```diff
@@ -215,6 +218,7 @@
                 <field name="parent_id" string="Activity"/>
                 <field name="date_start"/>
                 <field name="date_end"/>
+                <field name="asterisk_validate_record"/>
```
eilst, line 222: This should go at the end, i.e after  <field name="task_state"/> column

eilst (approved):

project_event/views/project_task_view.xml
jananjoy: @eilst : It is based on task description

PR #134, opened 2018-10-26
[ADD] No changes status done Task

eilst:


project_event/models/project_task.py
```diff
@@ -321,19 +321,21 @@ def draft_resources_reservation(self):
 
     @api.multi
     def action_cancel(self):
-        self.cancel_resources_reservation()
         if self.activity_task_type == 'task' and \
                 self.task_state in ['requested', 'read', 'postponed', 'accepted']:
             self.send_message('canceled')
+            self.cancel_resources_reservation()
+            self.write({'task_state': 'canceled'})
```
rimbendhaou, line 328: @eilst why cancel state of calendar event is cancelled with double l and for task_state is canceled with one l (it should have the same spelling)

rimbendhaou (changes requested):
@eilst when we have a new task and we execute the option action then request then accept it will create more then one calendar event for the task resource reservation. I got this problem when i added the unit test.

rimbendhaou (approved):

PR #137, opened 2018-10-26
[DEV] Behavior Top down of Activity to  automatically created Main Task  ( #281)

jananjoy:


project_event/models/project_task.py
```diff
@@ -183,6 +183,8 @@ def _check_subtask_project(self):
 
     @api.model
     def create(self, vals):
+
+        print("Testing create in project task ")
```
rimbendhaou: @jananjoy you shouldnt keep this code

project_event/models/project_task.py
```diff
@@ -215,6 +217,29 @@ def create_new_activity(self, vals):
             .next_by_code('project.task.activity')
         return super(Task, self).create(vals)
 
+    @api.multi
+    def write(self, vals):
+        if self.activity_task_type == 'activity':
+            self.write_task(vals)
+            self.write_activity(vals)
+        elif self.activity_task_type == 'task':
```
rimbendhaou, line 228: @jananjoy what if it is not task or activity? need another else

project_event/models/project_task.py
```diff
@@ -215,6 +217,29 @@ def create_new_activity(self, vals):
...
+
+    @api.multi
+    def write_task(self, vals):
+        main_task = self.get_main_task()
+        return main_task.write(vals)
+
+    def get_main_task(self):
+        return self.env['project.task'].search([
+            ('parent_id', '=', self.id),
+            ('name', '=', self.name)]
```
rimbendhaou: what if we change the name of the activity or the main task? better to distinguish the main task by a boolean field (is_main_task)

rimbendhaou: if we change the room id in main task by equipment and then we change the room id of the parent activity we will have a task with 2 resources (room and equipment) we must only have one resource per task

rimbendhaou (changes requested):

project_event/models/project_task.py
jananjoy: @rimbendhaou  Done

project_event/models/project_task.py
jananjoy: @rimbendhaou :Done

project_event/models/project_task.py
jananjoy: @rimbendhaou :+1:  I have added the field and change accordingly

rimbendhaou (approved):

PR #144, opened 2018-10-30
[ADD] Bookable validation

eilst:


project_event/models/project_task.py
```diff
@@ -175,6 +176,28 @@ def _onchange_resource_type(self):
         self.room_id = None
         self.equipment_id = None
 
+    @api.onchange('room_id')
+    def verify_room_bookable(self):
+        if self.room_id:
```
jananjoy, line 192: @eilstWhat is the validation message "if self.room_id; " condition fails Or if this condition is going to true always what is the use of checking this condition.

project_event/models/project_task.py
```diff
@@ -175,6 +176,28 @@ def _onchange_resource_type(self):
...
+                raise ValidationError(
+                        _(
+                            'This room is not bookable'
+                        )
+                    )
+                self.room_id = None
+
+    @api.onchange('equipment_id')
+    def verify_equipment_bookable(self):
+        if self.equipment_id:
```
jananjoy, line 201: @eilstWhat is the validation message "if self.equipment_id: " condition fails Or if this condition is going to true always what is the use of checking this condition.

project_event/models/project_task.py
```diff
@@ -175,6 +176,28 @@ def _onchange_resource_type(self):
...
+            if not self.equipment_id.is_bookable:
+                raise ValidationError(
+                        _(
+                            'This resource is not bookable'
+                        )
+                    )
+                self.equipment_id = None
```
jananjoy: @eilst: Better to move this statement   self.equipment_id = None before raising the validation error message

jananjoy (changes requested):
You have a few  change requests

project_event/models/project_task.py
eilst: this is a condition to catch the error, so the else condition is implicit, to pass

project_event/models/project_task.py
```diff
@@ -175,6 +176,26 @@ def _onchange_resource_type(self):
         self.room_id = None
         self.equipment_id = None
 
+    @api.onchange('room_id')
+    def verify_room_bookable(self):
```
rimbendhaou, line 191: @eilst def onchange_room_id

project_event/models/project_task.py
```diff
@@ -175,6 +176,26 @@ def _onchange_resource_type(self):
...
+        if self.room_id:
+            if not self.room_id.is_bookable:
+                raise ValidationError(
+                        _(
+                            'This room is not bookable'
+                        )
+                    )
+                
+    @api.onchange('equipment_id')
+    def verify_equipment_bookable(self):
```
rimbendhaou, line 200: @eilst def onchange_equipment_id

rimbendhaou (changes requested):

rimbendhaou (approved):

jananjoy (approved):

PR #163, opened 2018-11-06
[ADD] Use status  fields  in calendar event model

jananjoy:


project_resource_calendar/views/equipment_calendar_view.xml
```diff
@@ -4,50 +4,70 @@
         <field name="name">calendar.event.form</field>
         <field name="model">calendar.event</field>
         <field name="arch" type="xml">
+             <header>
+				<button name="action_draft" string="Unconfirmed" attrs="{'invisible': [('state', '=', 'draft')]}"
```
eilst: You need to add the actions i,e def action_draft():..

eilst: And the name of the button, I think is better to change it to:  Draft (a verb because it's an action)

eilst (approved):

PR #167, opened 2018-11-07
[FIX] False responsible as recipient partner_id

eilst:


lgharib (changes requested):
We shouldn't browse the first res.partner from the table because if the table is empty this code will break.
To handle the case of return none value use a Try Catch Exceptions mechanism.

lgharib (approved):

PR #169, opened 2018-11-07
[ADD] Added category to calendar event

lgharib:
Added category to calendar event


project_event/models/calendar_event.py
```diff
@@ -0,0 +1,22 @@
...
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
+from odoo import fields, models
+
+
+class CalendarEvent(models.Model):
+    _inherit = 'calendar.event'
+
+    def default_category_id(self):
+        return self.env['task.category'].search(
```
eilst, line 11: handle the case when there is no records in the table and maybe the reference would be more precise with a ref(rec_id

eilst (changes requested):

eilst (approved):

PR #177, opened 2018-11-09
[ADD] Historization of the 'Remarks' entry

jananjoy:


project_event/models/mail_thread.py
```diff
@@ -34,6 +34,9 @@ def _message_track(self, tracked_fields, initial):
             if col_name == 'user_id':
                 col_info['string'] = _('Created by')
 
+            if col_name == 'notes':
+                col_info['string'] = _('Note is created/modified')
```
eilst: "Note: "  or "Note was created/modified'"...apart from that, everything seems right here

eilst (changes requested):

project_event/models/mail_thread.py
jananjoy: @eilst :+1:  Done

eilst (approved):

PR #182, opened 2018-11-12
[Update] Activities tree view

eilst:


project_event/views/project_task_view.xml
```diff
@@ -97,6 +97,7 @@
             <tree string="Activity" default_order="date_start asc">
                 <field name="name"/>
                 <field name="code"/>
+                <field name="room_id"/>
```
jananjoy, line 100: Please check language conversion and field orders are correct based on the updated spec.

eilst: I did it, but maybe have a different one. Could you send me the spec that you think is right? Thank you

rimbendhaou (approved):

PR #183, opened 2018-11-12
[ADD] Set status to draft on copy of task model (A,T)

eilst:


project_event/models/project_task.py
```diff
@@ -559,3 +559,14 @@ def check_resource_booked(self):
             if len(overlaps_equipment) > 0:
                 return True
         return False
+
+    @api.multi
+    def copy(self, default=None):
+        if default is None:
```
jananjoy, line 565: if default is None:
                     default = {}
                      default['task_state'] = 'draft'
                      return super(Task, self).copy(default)
Explaination: If default is None:
    set default = {} 
check the condition 1 , check the condition 2 again. It is wasting execution time. so better to return the control imediately if it is None. (Line is more but execution is faster - trade off condition)

eilst: You mean if not default.get('name'): ? (as condition 2?)

eilst: Actually this is a base odoo method, I only added :
 default['task_state'] = 'draft'
but since I m overwriting, I added the whole default method

eilst: How would you put it?

jananjoy (approved):

PR #186, opened 2018-11-13
[FIX] Followers create

eilst:


eilst:
Invalid condition, better check for specific model use case: done in https://github.com/savoirfairelinux/project-addons/pull/187

lgharib (commented):
In line 17 we need to check in 'partner_id' exists in vals

lgharib (approved):

PR #192, opened 2018-11-14
[ADD] Deleting the default date when creating an activity

jananjoy:


eilst (changes requested):
To set the default date to None when creating an activity, you can extend the base odoo method. Adding a new field is definitely not ok. This field is really important. Creating a new field and replacing it on the xml will not solve the problem.

PR #198, opened 2018-11-15
[ADD] Default Task start date is empty

rimbendhaou:
Default Task start date is empty

### DIFF
--- a/project_event/models/project_task.py
+++ b/project_event/models/project_task.py
@@ -25,7 +25,7 @@ class Task(models.Model):
     )
     date_start = fields.Datetime(
         string='Starting Date',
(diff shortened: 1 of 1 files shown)


project_event/models/project_task.py
```diff
@@ -139,6 +139,12 @@ class Task(models.Model):
         default=False,
     )
     color = fields.Char(related='category_id.color')
+    date_start = fields.Datetime(
```
jananjoy: @rimbendhaou : Define two times "date_start" in the same model . Beside no need to redefine  date_end in project.task for this issue.

jananjoy (approved):

PR #199, opened 2018-11-15
[FIX] State of main task not directly linked to Activity #441

eilst:


project_event/models/project_task.py
```diff
@@ -280,7 +285,7 @@ def write_activity(self, vals):
         return super(Task, self).write(vals)
 
     @api.multi
-    def write_task(self, vals):
+    def write_main_task(self, vals):
```
jananjoy, line 283: @eilst: method name write_main_task () look good but main_task is not use  anywhere so you can write as 
return self.get_main_task().main_task.write(vals)

project_event/models/project_task.py
```diff
@@ -265,7 +265,12 @@ def create_new_activity(self, vals):
     @api.multi
     def write(self, vals):
         if self.activity_task_type == 'activity':
-            self.write_task(vals)
+            if 'task_state' in vals:
+                state = vals.pop('task_state')
```
jananjoy: @eilst : delete line 269 and change the line 271 as   vals['task_state'] = vals.pop('task_state').

project_event/models/project_task.py
eilst: @jananjoy True, you created that method, and I thought the same about just used once. But actually, since main task is managed in a different manner than a normal task, I thought its likely that we are going to refer this method later. So in my opinion it should stay, It makes the code more clear.

project_event/models/project_task.py
eilst: @neophilic_insect07  Done, I put this on the write_main_task method

rimbendhaou (approved):

PR #201, opened 2018-11-16
[FIX] FIX Event option action

rimbendhaou:


project_event/models/project_project.py
```diff
@@ -93,9 +93,29 @@ def action_accept(self):
     def action_option(self):
```
eilst, line 93: When we are already in state accepted anmd we return the event to draft, the wizard display this message:
Please Confirm your reservation.
The Following resources are already booked:
R1 
R1
Do you want to continue?

This message should only be displayed when we go from nothing reserved to reserved in option , but not to already reserved and accepted to reserved in option.

project_event/models/project_task.py
```diff
@@ -429,6 +433,22 @@ def open_resources_reservation(self):
             }
         )
 
+    @api.multi
+    def do_reservation(self):
```
eilst, line 448: def reserve(self): wouldn't be better?

eilst (changes requested):

eilst (approved):

PR #202, opened 2018-11-19
[ADD] Default category to A,T #399

eilst:


project_event/models/project_task.py
```diff
@@ -140,6 +141,11 @@ class Task(models.Model):
     )
     color = fields.Char(related='category_id.color')
 
+    def get_default_category(self):
+        if len(self.env['task.category'].search([('is_default', '=', True)])) > 1:
```
rimbendhaou: you can use limit=1 for search method to return only one record

project_event/models/project_task.py
```diff
@@ -39,6 +39,7 @@ class Task(models.Model):
     category_id = fields.Many2one(
         'task.category',
         string='Category',
+        default=lambda self: self.get_default_category()
```
rimbendhaou: You can put it directly : default=lambda self: self.env['task.category'].search([('is_default', '=', True)], limit=1)

rimbendhaou: I checked with @wbeverly we can use the boolean field is_default but we must display it in the form view (like that the admin can select the default category) and also we must add a constraint to only allow one true vale for the is_default field

project_event/models/task_category.py
```diff
@@ -18,3 +18,6 @@ class TaskCategory(models.Model):
         string='Sequence',
     )
     color = fields.Char()
+    is_default = fields.Boolean(
```
rimbendhaou, line 22: We have to display the field in the form view (for admin to select the default value)

project_event/data/task_categories.xml
```diff
@@ -1,7 +1,10 @@
 <?xml version="1.0" encoding="utf-8"?>
-<odoo noupdate="1">
+<odoo>
+    <data noupdate="1">
     <record id="task_category_uncategorized" model="task.category">
       <field name="name">Uncategorized</field>
       <field name="color">#FFFFFF</field>
+      <field name="is_default">True</field>
```
rimbendhaou, line 7: Better to put the default value selected manually by the admin

rimbendhaou (changes requested):

project_event/data/task_categories.xml
eilst: I`m loading this field as default. The user can change it later

project_event/models/task_category.py
eilst: Done

project_event/models/project_task.py
eilst: Done

project_event/models/task_category.py
```diff
@@ -18,3 +18,13 @@ class TaskCategory(models.Model):
...
+        string='Default',
+    )
+
+    @api.constrains('is_default')
+    def _change_is_default(self):
```
rimbendhaou, line 27: It is not a constraint. Constraint must check a condition

rimbendhaou (changes requested):

project_event/models/task_category.py
eilst: constraint on is_default: and changes is_default in all other existing records

rimbendhaou (approved):
Looks ok

PR #211, opened 2018-11-21
[ADD] Link calendar event to task

rimbendhaou:
Link calendar event to task


project_resource_calendar/views/equipment_calendar_view.xml
```diff
@@ -4,56 +4,96 @@
         <field name="name">calendar.event.form</field>
```
eilst, line 4: Just tested this view. And category appears unaligned, 
I did this:
you should try editing an existing record, or create one and then edited on the calendar view.
On editing popup, the category is not correctly displayed.

eilst (approved):

PR #215, opened 2018-11-22
[ADD] Added javascript saml error handler

lgharib:
Added javascript saml error handler

### DIFF
--- a/project_event/static/src/js/saml_error_handler.js
+++ b/project_event/static/src/js/saml_error_handler.js
@@ -1,0 +1,28 @@
+odoo.define('project_event.login_alert_message_saml', function (require) {
+    "use strict";
+
+
+
+
+function getUrlVars() {
+    var vars = {};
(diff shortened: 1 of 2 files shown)


project_event/static/src/js/saml_error_handler.js
```diff
@@ -0,0 +1,26 @@
...
+    var vars = {};
+    var parts = window.location.href.replace(/[?&]+([^=&]+)=([^&]*)/gi, function(m,key,value) {
+        vars[key] = value;
+    });
+    return vars;
+}
+
+switch(getUrlVars()["saml_error"]){
+    case "1":
+        $("#password").after("<br/><p class='alert alert-danger' >Il n'est pas autorisé de s'inscrire sur cette base de donnée.</p>");
```
eilst, line 15: Translations?
I think these messages should be in english, from the context we can get the language. You can check ihis link:
https://stackoverflow.com/questions/43892397/odoo-how-to-translate-strings-in-javascript-file

eilst (changes requested):

eilst (approved):

PR #216, opened 2018-11-22
[ADD] Weekly Reports #427

eilst:


project_resource_calendar/views/equipment_calendar_view.xml
```diff
@@ -88,9 +88,9 @@
         <field name="arch" type="xml">
            <xpath expr="//field[@name='message_needaction']" position="after">
                 <field name="state"/>
+                <field name="weekday"/>
```
rimbendhaou, line 131: we need to display this?

project_resource_calendar/report/calendar_event_report_view.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
+<odoo>
+    <report id="project_resource_calendar_report"
```
rimbendhaou, line 3: project_resource_calendar_weekly_event_report

rimbendhaou (changes requested):
the name of the room should appear only in the top (title) not in each line (according to @wbeverly ).
Better to reduce to margin size of the report

project_resource_calendar/views/equipment_calendar_view.xml
eilst: Yes

project_resource_calendar/report/calendar_event_report_view.xml
rimbendhaou: this is not a view better to rename the file to (calendar_event_weekly_report.xml)

project_resource_calendar/__manifest__.py
```diff
@@ -26,6 +26,9 @@
         'views/resource_view.xml',
         'views/room_view.xml',
         'views/instrument_view.xml',
+        'templates/assets.xml',
+        'report/calendar_event_report_view.xml',
```
rimbendhaou, line 30: Better to rename the file to calendar_event_weekly_report we will have also another report with this name in another PR

rimbendhaou (changes requested):

PR #218, opened 2018-11-23
[ADD] ADD confirmation for accept and request

rimbendhaou:
ADD confirmation for accept and request

### DIFF
--- a/project_event/models/project_project.py
+++ b/project_event/models/project_project.py
@@ -85,37 +85,13 @@ class Project(models.Model):
     def action_accept(self):
         if self.state in ['draft', 'option', 'postponed', 'canceled']:
             self.send_message('accepted')
-            for activity in self.task_ids:
-                activity.action_accept()
-        self.write({'state': 'accepted'})
+            return self.get_confirmation_wizard('accept')
 
     @api.multi
     def action_option(self):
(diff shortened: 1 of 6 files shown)


project_event/models/project_project.py
```diff
@@ -85,9 +85,30 @@ def action_cancel(self):
     def action_accept(self):
         if self.state in ['draft', 'option', 'postponed', 'canceled']:
             self.send_message('accepted')
+            res = ''
```
eilst: This piece of code is very similar to action_option, the only difference I see is on the message. Don't you think we could create a new method and call it with different messages?

project_event/models/project_task.py
```diff
@@ -340,12 +340,31 @@ def action_done(self):
 
     @api.multi
     def action_request(self):
-        self.draft_resources_reservation()
-        if self.activity_task_type == 'task' and \
-                self.task_state in ['draft', 'option', 'postponed', 'canceled']:
-            self.send_message('requested')
-        self.open_resources_reservation()
-        self.write({'task_state': 'requested'})
+        self.ensure_one()
```
eilst: Same thing here, this piece of code is used in more than one place

eilst (approved):

PR #221, opened 2018-11-26
[ADD] Added synchronisation between tasks and clones in calendar

lgharib:
Added synchronisation between tasks and clones in calendar

### DIFF
--- a/project_event/models/project_task.py
+++ b/project_event/models/project_task.py
@@ -286,7 +286,26 @@ class Task(models.Model):
                 task.write({'responsible_id': self.responsible_id.id,
                            'partner_id': self.partner_id.id})
         elif self.activity_task_type == 'task':
-            return super(Task, self).write(vals)
+            self.ensure_one()
+            update_task = super(Task, self).write(vals)
+            if update_task and self.reservation_event_id:
+                reservation_event = self.env['calendar.event'].browse(
+                    self.reservation_event_id)
+                reservation_event.write(
+                    {
+                        'start': self.date_start,
+                        'stop': self.date_end,
+                        'name': self.name,
+                        'resource_type': self.resource_type,
+                        'room_id': self.room_id.id if self.room_id else None,
+                        'equipment_ids': [(4, self.equipment_id.id,
+                                           0)] if self.equipment_id else None,
+                        'event_task_id': self.id,
+                        'category_id': self.category_id.id,
+                        'color': self.color,
+                    }
+                )
+            return update_task
         else:
             return super(Task, self).write(vals)
 
@@ -406,6 +425,8 @@ class Task(models.Model):
             'state': 'open',
             'event_task_id': self.id,
             'is_task_event': True,
+            'category_id': self.category_id.id,
+            'color': self.color,
         }
         new_event = calendar_event.create(values)
         self.reservation_event_id = new_event.id
--- a/project_event/models/project_task.py
+++ b/project_event/models/project_task.py
@@ -283,11 +283,33 @@ class Task(models.Model):
         return super(Task, self).create(vals)
 
     @api.multi
+    def update_reservation_event(self, update_task):
+        if update_task and self.reservation_event_id:
+            self.ensure_one()
+            reservation_event = self.env['calendar.event'].browse(
+                self.reservation_event_id)
+            reservation_event.write(
+                {
+                    'start': self.date_start,
+                    'stop': self.date_end,
+                    'name': self.name,
+                    'resource_type': self.resource_type,
+                    'room_id': self.room_id.id if self.room_id else None,
+                    'equipment_ids': [(6, 0,
+                                       self.get_equipment_ids_inside())] if self.room_id
+                    else [(6, 0, self.equipment_id.id)],
+                    'category_id': self.category_id.id,
+                }
+            )
+
+    @api.multi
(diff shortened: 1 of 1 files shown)


project_event/models/project_task.py
```diff
@@ -286,7 +286,26 @@ def write(self, vals):
...
+                reservation_event.write(
+                    {
+                        'start': self.date_start,
+                        'stop': self.date_end,
+                        'name': self.name,
+                        'resource_type': self.resource_type,
+                        'room_id': self.room_id.id if self.room_id else None,
+                        'equipment_ids': [(4, self.equipment_id.id,
+                                           0)] if self.equipment_id else None,
+                        'event_task_id': self.id,
```
rimbendhaou: @lgharib event_task_id will not be changed, no need to add it to the write method

project_event/models/project_task.py
```diff
@@ -286,7 +286,26 @@ def write(self, vals):
...
+                reservation_event = self.env['calendar.event'].browse(
+                    self.reservation_event_id)
+                reservation_event.write(
+                    {
+                        'start': self.date_start,
+                        'stop': self.date_end,
+                        'name': self.name,
+                        'resource_type': self.resource_type,
+                        'room_id': self.room_id.id if self.room_id else None,
+                        'equipment_ids': [(4, self.equipment_id.id,
```
rimbendhaou: For equipment_ids we have to reserve all the equipment inside a selected room_id. so we can have more than one equipment in the event.

rimbendhaou (changes requested):
wait for spec validation about the equipment ids

project_event/models/project_task.py
```diff
@@ -286,7 +286,26 @@ def write(self, vals):
                 task.write({'responsible_id': self.responsible_id.id,
                            'partner_id': self.partner_id.id})
         elif self.activity_task_type == 'task':
-            return super(Task, self).write(vals)
+            self.ensure_one()
```
eilst: If you want to ensure we have a singleton, this self.ensure_one() should go at the beginning of the method, not here. But, what would be better is to create a new method, because the write method is already overwhelmed. it does too many things. Better to put your feature in a new method and use it when writing

jananjoy: @lgharib This method refactor completely so you need to test and change your code accordingly

project_event/models/project_task.py
```diff
@@ -282,12 +282,43 @@ def create_new_activity(self, vals):
...
+            )
+
     @api.multi
     def write(self, vals):
         if self.activity_task_type == 'activity':
             return self.write_activity(vals)
         else:
-            return super(Task, self).write(vals)
+            update_task = super(Task, self).write(vals)
+            self.update_reservation_event(update_task)
```
eilst: why you pass update_task? I think it should be either self since self=update_task or vals if you want to pass only the changed values

project_event/models/project_task.py
```diff
@@ -282,12 +282,43 @@ def create_new_activity(self, vals):
             .next_by_code('project.task.activity')
         return super(Task, self).create(vals)
 
+    @api.multi
+    def update_reservation_event(self, update_task):
+        if update_task and self.reservation_event_id:
```
eilst: self.ensur_one at the beginning. rather than using update_task, you can use self

project_event/models/project_task.py
```diff
@@ -282,12 +282,43 @@ def create_new_activity(self, vals):
...
+                    partners.append(e.user_id.partner_id.id)
+                else:
+                    raise UserError(
+                        _('Please define user account for the %s employee') % (
+                            e.name,))
+            reservation_event.write(
+                {
+                    'start': self.date_start,
+                    'stop': self.date_end,
+                    'name': self.name,
```
eilst: only start and stop have different keys (vs date_start and date_end), for the others you could pop these  if present and then dictionary.update(vals) and then write

eilst (changes requested):

project_event/models/project_task.py
```diff
@@ -282,12 +282,43 @@ def create_new_activity(self, vals):
...
+                    'partner_ids': [(6, 0, partners)],
+                    'category_id': self.category_id.id,
+                }
+            )
+
     @api.multi
     def write(self, vals):
         if self.activity_task_type == 'activity':
             return self.write_activity(vals)
         else:
```
rimbendhaou, line 322: This else means if activity_task_type is task or if it is a project task (the module project management created by default in odoo)

project_event/models/project_task.py
```diff
@@ -282,11 +282,50 @@ def create_new_activity(self, vals):
...
         return super(Task, self).create(vals)
 
+    @api.multi
+    def get_partners(self):
+        partners = []
+        for e in self.employee_ids:
+            if e.user_id:
+                partners.append(e.user_id.partner_id.id)
+            else:
+                raise UserError(
```
rimbendhaou: @lgharib the raise error is not handled when updating task employee_ids

rimbendhaou (changes requested):

eilst (approved):

project_event/models/project_task.py
eilst: True

eilst (changes requested):

PR #224, opened 2018-11-28
[FIX] Write method on task (enhacement + fix: tasks resposible only changes when activity resp changes)

eilst:


rimbendhaou (commented):
better to add unit test to check that the top to down modification is done successfully and the write function behavior is correct

rimbendhaou (approved):

PR #227, opened 2018-11-29
[ADD] Added filters by category to calendars

lgharib:
Added filters by category to calendars


rimbendhaou (approved):
It works only after page refresh

PR #240, opened 2018-12-04
[FIX] Log notes

eilst:


project_event/models/mail_thread.py
```diff
@@ -35,7 +35,8 @@ def _message_track(self, tracked_fields, initial):
                 col_info['string'] = _('Created by')
 
             if col_name == 'notes':
-                col_info['string'] = _('Note was created/modified')
+                col_info['string'] = _('Note')
+                new_value = 'created/modified'
```
jananjoy, line 39: @eilst  : What is the purpose? In there any anomaly or change request for this task.

project_event/models/mail_tracking_value.py
```diff
@@ -14,12 +14,16 @@ class MailTracking(models.Model):
...
+        if col_info['type'] in ['integer', 'float', 'char', 'text', 'datetime', 'monetary']:
             values.update({
                 'old_value_%s' % col_info['type']: initial_value,
                 'new_value_%s' % col_info['type']: new_value
             })
+        elif col_info['type'] == 'html':
+            col_info['type'] = 'char'
+            values.update({
+                'new_value_%s' % col_info['type']: new_value
+            })
```
jananjoy, line 26: @eilst if this is for refactoring purpose, It sounds good. otherwise  what problem this new code  solve?

eilst: yes, the message was wrong, there is a specific way an html should be handled on the log messages. It was not done correctly. I corrected this.

jananjoy: Cool as per the task specification,  It is  implementated accordingly, This is just a message string. However good work.

project_event/models/mail_thread.py
```diff
@@ -35,7 +35,8 @@ def _message_track(self, tracked_fields, initial):
                 col_info['string'] = _('Created by')
 
             if col_name == 'notes':
-                col_info['string'] = _('Note was created/modified')
+                col_info['string'] = _('Note')
```
rimbendhaou, line 38: @eilst what if we have a note field with type char?

rimbendhaou: @eilst after functional test, even if we mofify the description we have the notification : Note: created/modified

rimbendhaou (changes requested):

rimbendhaou (approved):

PR #244, opened 2018-12-06
[ADD] Display modified fields in the chatter

rimbendhaou:
Display modified fields on event, activity and task


project_event/models/mail_thread.py
```diff
@@ -25,14 +25,6 @@ def _message_track(self, tracked_fields, initial):
             track_visibility = getattr(self._fields[col_name], 'track_visibility', 'onchange')
             initial_value = initial[col_name]
             new_value = getattr(self, col_name)
```
eilst, line 27: I do remember the spec for the message on the field employee_ids (The assign people). You are removing the message Assigned to and created by and letting the default odoo values. You are changing this to be handled in mail_tracking_value.py. But even if you are solving the case in which you don't want to format the message of field with the same name, what you are doing now is formating the messages of fields of the same type. So the problem is now more general.

project_event/models/mail_tracking_value.py
```diff
@@ -46,6 +46,13 @@ def create_tracking_values(self, initial_value, new_value, col_name, col_info):
                 'old_value_char': initial_value and initial_value.name_get()[0][1] or '',
                 'new_value_char': new_value and new_value.name_get()[0][1] or ''
             })
+        elif col_info['type'] == 'many2many':
+            values.update({
+                'old_value_char': initial_value and list(
+                    map(lambda x: x.name, initial_value)) or '',
+                'new_value_char': new_value and list(
```
eilst, line 53: ', '.join(list(map(lambda employee: employee.name, new_value))) need to use this one; otherwise, you are passing a list with ''['  Employees   "]"". We dont want the square brackets

eilst (changes requested):

project_event/models/mail_thread.py
eilst: One possible solution is to check for the res.model to be project.task or project.project so the changes are applied only to these two models.

eilst (approved):

PR #248, opened 2018-12-07
[FIX] Preceding Succeding Visibility & Done status bar

eilst:


project_event/models/project_task.py
```diff
@@ -130,6 +130,12 @@ class Task(models.Model):
         default='draft',
         track_visibility='onchange',
     )
+    task_state_report_done_required = fields.Selection(
+        related='task_state'
+    )
+    task_state_report_not_done_required = fields.Selection(
```
rimbendhaou, line 136: why two similar fields? what is  value difference between these fields

rimbendhaou (commented):
why two similar fields? what is value difference between these fields

project_event/models/project_task.py
eilst: For the display on the form view, with different attributes. If the filters are applied to the same field, it takes the last definition of the filter. For this attribute:
statusbar_visible="draft,requested,read,accepted,approved,postponed,canceled"/>
It takes the last definition, even if only one of the two is displayed

rimbendhaou (approved):

PR #264, opened 2018-12-17
[FIX] Update Calendar Event Form and Pop up

eilst:


project_resource_calendar/models/calendar_event.py
```diff
@@ -49,6 +49,18 @@ class CalendarEvent(models.Model):
...
+        compute='_compute_dates',
+        inverse='_inverse_dates',
+        store=True,
+        states={'done': [('readonly', [('id', '!=', False), ('recurrency','=',True), ('is_task_event','=',True)])]},
+        track_visibility='onchange',
+        )
+    duration = fields.Float(
+        'Duration',
+        states={'done': [('readonly', [('id', '!=', False), ('recurrency','=',True), ('is_task_event','=',True)])]})
+
```
jananjoy, line 63: @eilst What is the technical advantages of moving code from front-end validation to back-end validation.

lgharib (approved):

project_resource_calendar/models/calendar_event.py
eilst: It was not working on the front end, I m not sure why. But it seems that the XPath was not inserting the values specified on the front end because these two fields had the attribute read-only set to True on the back end. So even if it is specified something on the front end, if it's blocked on the back end it is not going to work. That's why the change needed to be done on the back end.

PR #278, opened 2018-12-26
[FIX] Sector Field on a Calendar.event should not be mandatory

jananjoy:


lgharib (approved):

project_event/views/calendar_event_category_view.xml
```diff
@@ -13,7 +13,7 @@
                 </h2>
                 <label for="sector_id" class="oe_edit_only"/>
                 <h2>
-                    <field name="sector_id" attrs="{'readonly': [('is_task_event','=',True)], 'required': True}"/>
+                    <field name="sector_id" attrs="{'readonly': [('is_task_event','=',True)], 'required': False}"/>
```
rimbendhaou: you can simply delete the required attribute because by default it is False

jananjoy: @rimbendhaou Done

PR #297, opened 2019-01-08
[FIX]  Refactoring  is_new_activity() and  is_new_task  in project.task

jananjoy:


project_event/models/project_task.py
```diff
@@ -329,13 +329,15 @@ def create_main_task(self, vals, parent_id):
 
     def is_new_task(self, vals):
         if 'activity_task_type' in vals:
-            return vals['activity_task_type'] == 'task'
+            vals['activity_task_type'] == 'task'
+            return True
```
rimbendhaou: It will always return true in this case. we dont need to refactor this code

rimbendhaou (changes requested):

project_event/models/project_task.py
jananjoy: @rimbendhaou  I know but method signature has to be consistent e.g.
def isEmpty(self, dictionary):
   for element in dictionary:
     if element:
       return True
     return False

eilst (approved):

rimbendhaou (approved):

PR #302, opened 2019-01-09
[ADD] Pricing to project_task from resources

eilst:


project_event/models/project_task.py
```diff
@@ -206,6 +206,25 @@ class Task(models.Model):
...
+            if rec.room_id:    
+                rec.pricing = str(rec.room_id.pricing) \
+                              + " / " \
+                              + rec.pricing_type
+            elif rec.equipment_id:
+                rec.pricing = str(rec.equipment_id.pricing) \
+                              + " / " \
+                              + rec.pricing_type
+            else:
+                rec.pricing = '0'
```
jananjoy, line 227: @eilst : When this condition will happen? write the test case for this new method. how can test this functionality manually?

eilst:
This functionality exists already by selecting 'Room' -> 'Pricing'. This PR puts pricing directly on the field list, but this is no longer needed.

PR #304, opened 2019-01-09
[FIX] Message sent when status changed

eilst:


project_event/models/project_project.py
```diff
@@ -104,14 +104,10 @@ def action_cancel(self):
 
     @api.multi
     def action_accept(self):
-        if self.state in ['draft', 'option', 'postponed', 'canceled']:
-            self.send_message('accepted')
-            return self.get_confirmation_wizard('accept')
+        return self.get_confirmation_wizard('accept')
```
jananjoy, line 107: @eilst If this is the case, we don't require def action_accept(self): we can directly call with  self.get_confirmation_wizard('accept') . Correct me if i am wrong

project_event/models/project_project.py
```diff
@@ -104,14 +104,10 @@ def action_cancel(self):
...
 
     @api.multi
     def action_option(self):
-        if self.state == 'accepted':
-            self.send_message('option')
         return self.get_confirmation_wizard('option')
```
jananjoy, line 111: @eilst ; Same comment as above.

project_event/models/project_project.py
```diff
@@ -154,6 +150,8 @@ def confirm_accept_reservation(self):
                 activity.send_message('requested')
             activity.open_resources_reservation()
             activity.write({'task_state': 'accepted'})
+        if self.state in ['draft', 'option', 'postponed', 'canceled']:
+            self.send_message('accepted')
```
jananjoy, line 154: @eilst  send_message() have only statement  self.env['mail.message'].create(self.get_message(action)) . Why you do not think about generic approach for all project.task which have the similar  syntax and format of referiing self.env[]

rimbendhaou (approved):

PR #305, opened 2019-01-10
[DEV] Test case for validating "recurrent_state" and "recurrent_type" in calendar event model 

jananjoy:


project_resource_calendar/tests/test_calendar_event.py
```diff
@@ -42,3 +42,36 @@ def test_010_onchange_room_id(self):
         self.calendar_event._onchange_room_id()
         self.assertEqual(self.calendar_event.room_id.name,
                          'Test Room id after onchange method execution')
+
+    def test_020_calculate_recurrent(self):
+        self.recurrency = True
```
rimbendhaou: the recurrent_state field should be calculated automatically when recurrency field is updated

project_resource_calendar/tests/test_calendar_event.py
```diff
@@ -42,3 +42,36 @@ def test_010_onchange_room_id(self):
...
+
+        self.recurrency = False
+        if self.recurrency:
+            self.recurrent_state = "Yes"
+        else:
+            self.recurrent_state = "No"
+        self.assertEqual(self.recurrent_state, 'No')
+
+    def test_030_calculate_recurrence_type(self):
+        self.recurrency = True
```
rimbendhaou, line 53: the same comment

rimbendhaou (changes requested):

project_resource_calendar/tests/test_calendar_event.py
jananjoy: @rimbendhaou: But for testing, I purposely set the value for testing both the scenarios otherwise what is your comment to change in the existing code.

rimbendhaou:
Need to check the new added methods

eilst (approved):

PR #307, opened 2019-01-11
[FIX] Fix for calendar event form view

jananjoy:


project_resource_calendar/models/calendar_event.py
```diff
@@ -92,15 +92,22 @@ def _calculate_recurrent(self):
 
     def _calculate_recurrence_type(self):
         if self.recurrency:
+            if self.rrule_type == 'daily':
```
rimbendhaou: if self.rrule_type is empty?

rimbendhaou (changes requested):

project_resource_calendar/views/equipment_calendar_view.xml
```diff
@@ -112,8 +112,8 @@
                 <attribute name="invisible">1</attribute>
             </xpath>
             <xpath expr="//field[@name='location']" position="before">
-                <field name="recurrent_state" readonly="1"/>
-                <field name="recurrence_type" readonly="1"/>
+                <field name="recurrent_state" attrs="{'invisible': [('is_task_event','=',True)]}"/>
```
rimbendhaou: why you delete the readonly attributes

rimbendhaou (changes requested):

rimbendhaou (approved):

PR #312, opened 2019-01-14
[FIX] project.task for code optimization

jananjoy:


project_event/models/project_task.py
```diff
@@ -592,45 +593,41 @@ def reserve_equipment_inside(self, event_id):
...
-                'state': 'draft'
-            }
-        )
+            reserve_event = self.info_calendar_event()
+            reserve_event.write(
+                {'state': 'draft'}
+            )
 
     @api.multi
     def open_resources_reservation(self):
```
eilst: Looks good, you can improve it more by merging those three methods into one. "def change_resource_reservation(self, status_change)"

jananjoy: @eilst  Change according to your comment. Thx you

project_event/models/project_task.py
```diff
@@ -652,7 +632,7 @@ def action_cancel(self):
         if self.activity_task_type == 'task' and \
                 self.task_state in ['requested', 'read', 'postponed', 'accepted']:
             self.send_message('canceled')
-            self.cancel_resources_reservation()
+            self.write_resources_reservation_state('canceled')
```
eilst: it's 'cancelled', not 'canceled'

jananjoy: @eilst  done

rimbendhaou (approved):

PR #332, opened 2019-02-04
[FIX] Reference to weekday in different language than english

eilst:


project_resource_calendar/wizard/weekly_report.py
```diff
@@ -33,7 +34,7 @@ def format_event_to_docs(self, events, docs):
                 'name': event.name,
                 'start': event.start,
                 'stop': event.stop,
-                'weekday': event.weekday,
```
rimbendhaou, line 36: Do you still use the old weekday field? If no, you have to remove it from the model

rimbendhaou (changes requested):

project_resource_calendar/wizard/weekly_report.py
eilst: Right, I remove it with the now useless method _get_weekday and its references

rimbendhaou (approved):

PR #343, opened 2019-02-11
[ADD] client to calendar event form and required

eilst:


project_resource_calendar/models/calendar_event.py
```diff
@@ -179,3 +185,13 @@ def print_calendar_report(self):
         return self.env.ref(
             'project_resource_calendar.calendar_event_report'
         ).report_action(self)
+
+    @api.model
+    def create(self, vals):
```
rimbendhaou, line 199: @eilst what if we modify (we use write method) and we delete the client from the participants list? we will not do this verification(verify_client_in_participants)?

rimbendhaou (changes requested):

project_resource_calendar/models/calendar_event.py
eilst: Done, right, the validation should also be done when writing

project_event/models/calendar_event.py
```diff
@@ -21,3 +21,17 @@ class CalendarEvent(models.Model):
...
+        store=True,
+    )
+
+    @api.one
+    @api.depends('client_id')
+    def _get_client_tag_id(self):
+        if self.client_id:
+            self.client_tag = self.client_id.tag_id
+        else:
+            self.client_tag = None
```
rimbendhaou: PEP8 new line after None

project_event/models/calendar_event.py
```diff
@@ -21,3 +21,17 @@ class CalendarEvent(models.Model):
         string='Client Type',
     )
     color = fields.Char(related='category_id.color')
+    client_tag = fields.Many2one(
```
rimbendhaou, line 24: If it is a readonly field it is better to put it as a related field

eilst: Right, done

project_event/models/calendar_event.py
eilst: Done

rimbendhaou (approved):

PR #345, opened 2019-02-15
[ADD] Added client_type inheritance

lgharib:
Added client_type inheritance

### DIFF
--- a/project_event/models/project_task.py
+++ b/project_event/models/project_task.py
@@ -236,6 +236,8 @@ class Task(models.Model):
         if self.project_id:
             if self.project_id.responsible_id:
                 self.responsible_id = self.project_id.responsible_id
+            self.partner_id = self.project_id.partner_id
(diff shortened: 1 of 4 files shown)


project_event/models/project_task.py
```diff
@@ -354,6 +365,8 @@ def create_activity(self, vals):
             .next_by_code('project.task.activity')
         children = self.activity_has_children(vals)
         new_activity = super(Task, self).create(vals)
+        if self.project_id:
```
rimbendhaou: @lgharib this code will change the edited client_type with the client_type of the partner of the project_id

rimbendhaou (changes requested):

project_event/models/project_task.py
lgharib: Fixed

rimbendhaou (approved):

PR #347, opened 2019-02-22
[ADD] Added clone task changes status and child task

lgharib:
Added clone task changes status and child task

### DIFF
--- a/project_event/models/project_task.py
+++ b/project_event/models/project_task.py
@@ -667,26 +667,35 @@ class Task(models.Model):
         return self.env['calendar.event'].\
             browse(self.reservation_event_id)
 
+    def do_clone_task_reservation(self):
+        if self.reservation_event_id:
+            self.get_calendar_event().write({'state': 'draft'})
+
+    def do_task_reservation(self):
+        self.draft_resources_reservation()
+        if self.task_state not in ['option', 'done']:
+            self.send_message('option')
+        self.write({'task_state': 'option'})
+        self.do_clone_task_reservation()
+
     @api.multi
     def do_reservation(self):
         self.ensure_one()
-        if self.activity_task_type == 'task':
-            self.draft_resources_reservation()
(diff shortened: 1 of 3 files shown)


project_event/models/project_task.py
```diff
@@ -657,20 +657,28 @@ def info_calendar_event(self):
...
-                self.send_message('option')
         if self.is_activity():
             for child in self.child_ids:
-                child.draft_resources_reservation()
-                if child.task_state not in ['option', 'done']:
-                    child.send_message('option')
-                child.write({'task_state': 'option'})
+                child.do_task_reservation()
             self.send_message('option')
+            self.write({'task_state': 'option'})
```
eilst, line 714: Add a test for the new features (update clone status)

project_event/models/project_task.py
```diff
@@ -657,20 +657,28 @@ def info_calendar_event(self):
...
+            return
```
eilst: Since you already using an if statement and the function is not returning anything, an else seems better in this case. Don't you think?

project_event/models/project_task.py
```diff
@@ -657,20 +657,28 @@ def info_calendar_event(self):
         return self.env['calendar.event'].\
             browse(self.reservation_event_id)
 
+    def do_clone_task_reservation(self):
```
rimbendhaou, line 696: @lgharib I think that no need to execute get_confirmation_wizard. If we are in a request state we must change the state to option and change the equivalent calendar.event state to not confirmed

rimbendhaou (changes requested):

project_event/models/project_task.py
lgharib: I add a test to check that the status of the clone calendar task has changed after do_reservation

rimbendhaou (changes requested):
check test fails

rimbendhaou (approved):

PR #350, opened 2019-02-25
[ADD]Evaluate the behaviour of the 'Activité' Widget on top right hand corner in Odoo

jananjoy:
…


eilst (commented):
Check models.py for returning the right view (for project.project model and task model, not default views )

rimbendhaou (changes requested):
It doesnt work

eilst:
The requirement change and we will close this PR.

PR #351, opened 2019-02-25
[ADD] (Correction Feature) Add 'Spectators' entry / Adding 'Spectators' entry

jananjoy:


project_event/models/project_task.py
```diff
@@ -205,6 +205,11 @@ class Task(models.Model):
         string='Is Created From Template',
         default=False,
     )
+    spectators = fields.Integer(
+        string='Spectators',
+        required=False,
+        readonly=False,
```
eilst: I readonly= False, you don't need to specify it

jananjoy: @eilst : It is resolved

project_event/models/project_task.py
```diff
@@ -205,6 +205,16 @@ class Task(models.Model):
...
+    spectators = fields.Char(
+        string='Spectators',
+     )
+
+    @api.onchange('spectators')
+    def check_numeric(self):
+        try:
+            int(self.spectators)
+        except Exception as exception:
+            raise ValidationError('The Input value should be numeric')
```
eilst: raise ValidationError(
                        _(
                            'For Translations'
                        ) )

eilst (approved):

PR #359, opened 2019-03-04
[ADD] Removed acl for calendar timeline calendar event

lgharib:
Removed acl for calendar timeline calendar event


eilst:
This PR fixes the ACL problem for base users. But, it had a reason for being in the first place, to limit the write of calendar events (I think especially on the Timeline view, isn't?) So, if we can add a test case to verify this behavior and the necessary modifications,  we will be sure that it's ok.

lgharib:
The non editable state for is_task_event event in calendar event is handled in front end :
See file git/sfl/addons/project_specific/static/src/js/web_calendar_colour.js
                        if(event.record['is_task_event'] === true){
                            event.editable = false;
                        }

eilst (approved):

PR #363, opened 2019-03-05
[FIX] Double booking and client_id fix

rimbendhaou:
Double booking and client_id not required


project_resource_calendar/models/calendar_event.py
```diff
@@ -161,27 +160,27 @@ def _check_resources_double_book(self):
...
-            for resource in overlaps.mapped(lambda s: s.equipment_ids):
-                if resource.id in record.equipment_ids.ids:
-                    raise ValidationError(
-                        _(
-                            'The resource, %s, cannot be double-booked '
-                            'with any overlapping meetings or events.',
-                        ) % resource.name,
-                    )
+            for event in events:
+                if (event.start < record.stop) & (event.stop > record.start):
```
eilst: The logic is good. but maybe if you create a method:
def record_intersects_event(self, record, event):
    if (event.start < record.stop) & (event.stop > record.start):
        return True
    else:
        return False

and call this method in place, it's more readable, don't you think?

eilst (approved):

PR #366, opened 2019-03-05
[FIX] Set Resource type field invisible if it is not created …

lgharib:
…from a task"


project_event/tests/test_project_event_task.py
```diff
@@ -429,3 +429,17 @@ def test_160_check_partner_id_calendar_event_as_client_of_task(self):
             calendar_event.client_id.id,
             self.partner_1.id
         )
+    def test_170_update_task_event_clone_partner_ids_with_employee_ids(self):
+        import pdb
```
eilst: remove pdb

rimbendhaou (approved):

PR #374, opened 2019-03-08
[FIX] Font size client_id, room_id, resource_ids, partner_ids on calendar 

eilst:


rimbendhaou (changes requested):
It can be better!
Check with functional, if it is ok, it will be ok for me

eilst:
@rimbendhaou  How do you think it can be better?
I agree it can be better, I should completely rewrite this view since it contains 42 instances of xpath, which makes it incredibly confusing and hard to read.

rimbendhaou (approved):

PR #376, opened 2019-03-11
[FIX] Language Translation

eilst:


jananjoy (approved):
Check with language translator. It looks good

PR #396, opened 2019-03-21
[FIX]Error while replicating an Activities

jananjoy:


project_event/models/project_task.py
```diff
@@ -396,20 +396,26 @@ def create_children_from_activity_create(self, children, parent_id):
 
     @api.multi
     def write(self, vals):
-        if self.is_activity():
-            return self.write_activity(vals)
-        else:
-            self.update_reservation_event(vals)
-            return super(Task, self).write(vals)
+        super(Task, self).write(vals)
```
rimbendhaou: It will write 2 times?

project_event/models/project_task.py
```diff
@@ -396,20 +396,26 @@ def create_children_from_activity_create(self, children, parent_id):
...
+        for rec in self:
+            if rec.is_activity():
+                return rec.write_activity(vals)
+            else:
+                rec.update_reservation_event(vals)
+                return super(Task, rec).write(vals)
 
     def is_activity(self):
-        return self.activity_task_type == 'activity'
+        self.ensure_one()
```
rimbendhaou: if it is only one record why we need to use for rec in self?

rimbendhaou (changes requested):

project_event/models/project_task.py
jananjoy: @rimbendhaou  Changed Done

project_event/models/project_task.py
jananjoy: Remove the duplicate right  with update_reservation_events because values are set to vals in value return. I call super write() method one time . I also remove  super(Task, self).write(vals) from write_activity() method

project_event/models/project_task.py
```diff
@@ -401,46 +401,44 @@ def create_children_from_activity_create(self, children, parent_id):
 
     @api.multi
     def write(self, vals):
-        if self.is_activity():
-            return self.write_activity(vals)
-        else:
-            self.update_reservation_event(vals)
-            return super(Task, self).write(vals)
+        super(Task, self).write(vals)
```
eilst: You should return super(Task, self).write(vals)

jananjoy: what is the implication?

rimbendhaou (changes requested):
It raises exception when duplicate event (This task must have a project since they are linked to timesheets)

project_event/models/project_task.py
```diff
@@ -459,46 +459,45 @@ def create_children_from_activity_create(self, children, parent_id):
 
     @api.multi
     def write(self, vals):
-        if self.is_activity():
-            return self.write_activity(vals)
-        else:
-            self.update_reservation_event(vals)
-            return super(Task, self).write(vals)
+        update = super(Task, self).write(vals)
```
eilst: this  will cause doubleright

eilst (changes requested):

project_event/tests/test_project_event.py
```diff
@@ -46,3 +46,14 @@ def test_030_onchange_partner_id(self):
         self.project_1.partner_id = self.partner_tag2.id
         self.project_1._onchange_partner_id()
         self.assertEqual(self.project_1.client_type.name, 'Type 3')
+
+    def test_031_duplicate_activities(self):
+
```
eilst, line 51: remove line

project_event/models/project_task.py
```diff
@@ -459,46 +459,53 @@ def create_children_from_activity_create(self, children, parent_id):
...
-        if self.is_activity():
-            return self.write_activity(vals)
-        else:
-            self.update_reservation_event(vals)
-            return super(Task, self).write(vals)
+        update = None
+        for rec in self:
+            if rec.project_id:
+                update = super(Task, rec).write(vals)
+        for rec in self:
```
eilst, line 466: why 2 for loops on the same object? it can be put in one

project_event/models/project_task.py
```diff
@@ -459,46 +459,53 @@ def create_children_from_activity_create(self, children, parent_id):
...
             vals['child_ids'] = temp
         return main_task
 
+    def write_activty_children(self, vals):
+        activity_child_vals = {}
+        if 'responsible_id' in vals:
+            activity_child_vals['responsible_id'] = vals['responsible_id']
+        if 'partner_id' in vals:
+            activity_child_vals['partner_id'] = vals['partner_id']
+
```
eilst, line 502: Remove space. This method was renamed write_children->write_activty_children 
 and the variable task_vals -> activity_child_vals. It's ok, but since you are making it more explicit, you should name activity_child_vals -> activity_childen_vals

project_event/models/project_task.py
```diff
@@ -459,46 +459,53 @@ def create_children_from_activity_create(self, children, parent_id):
...
     @api.multi
     def write(self, vals):
-        if self.is_activity():
-            return self.write_activity(vals)
-        else:
-            self.update_reservation_event(vals)
-            return super(Task, self).write(vals)
+        update = None
+        for rec in self:
+            if rec.project_id:
```
eilst, line 464: If I understand the main changes of this PR are: 
for loop on write records:
and the checking of rec.project_id.
Right?

Could you please describe what was fixed? Normally it's ok but here I don't quite get it.  I think I more or less get it, but I'm also thinking that why check for project_id?

project_event/models/project_task.py
```diff
@@ -459,46 +459,53 @@ def create_children_from_activity_create(self, children, parent_id):
...
-        if 'partner_id' in vals:
-            task_vals['partner_id'] = vals['partner_id']
-        for task in self.child_ids:
-            if task == self.get_main_task():
-                continue
-            if task_vals:
-                task.write(task_vals)
-
     @api.multi
-    def write_main_task(self, vals):
```
eilst, line 490: I see that you renamed write_main_task -> write_activity_main_task.
Ok, for me, I think this method didn't require any modification. But it's ok if you think we need to be more explicit.

project_event/models/project_task.py
```diff
@@ -459,46 +459,53 @@ def create_children_from_activity_create(self, children, parent_id):
...
-        main_task = self.get_main_task()
+    def write_activity_main_task(self, vals):
         temp = []
         if 'task_state' in vals:
             return False
         if 'child_ids' in vals:
             temp = vals.pop('child_ids')
+        main_task = self.get_main_task()
         main_task = main_task.write(vals)
```
eilst, line 491: this can be improved: main_task = self.get_main_task().write(vals)

project_event/models/project_task.py
```diff
@@ -947,6 +955,7 @@ def is_resource_booked(self):
 
     @api.multi
     def copy(self, default=None):
+        print('project task: def copy')
```
eilst, line 958: remove print

project_event/models/project_task.py
```diff
@@ -512,6 +519,7 @@ def get_parent_project_id(self):
 
     @api.multi
     def update_reservation_event(self, vals):
+        self.ensure_one()
         if len(self) == 1:
```
eilst, line 523: len(self) == 1 is the same thing as self.ensure_one(). Right?

project_event/tests/test_project_event.py
```diff
@@ -46,3 +46,14 @@ def test_030_onchange_partner_id(self):
         self.project_1.partner_id = self.partner_tag2.id
         self.project_1._onchange_partner_id()
         self.assertEqual(self.project_1.client_type.name, 'Type 3')
+
+    def test_031_duplicate_activities(self):
```
eilst, line 50: Please follow agreements, the naming convention: test_040_du...

project_event/tests/test_project_event.py
```diff
@@ -46,3 +46,14 @@ def test_030_onchange_partner_id(self):
...
+
+        new_rec = self.test_duplicate_project_1.copy()
+        self.assertEqual(new_rec.name, 'Test Duplicate Project 1 (copy)')
+        self.assertEqual(
+            len(self.test_duplicate_activity_1.child_ids), 1
+        )
+        self.assertEqual(
+            len(self.test_duplicate_activity_2.child_ids), 1
```
eilst, line 58: Test for two, because you are not testing what you implemented. If I understand well, your changes are supposed to fix the problem: 1 event with 2 activities with multiple tasks. So test this.

project_event/tests/test_project_event.py
```diff
@@ -46,3 +46,14 @@ def test_030_onchange_partner_id(self):
         self.project_1.partner_id = self.partner_tag2.id
         self.project_1._onchange_partner_id()
         self.assertEqual(self.project_1.client_type.name, 'Type 3')
+
+    def test_031_duplicate_activities(self):
+
+        new_rec = self.test_duplicate_project_1.copy()
+        self.assertEqual(new_rec.name, 'Test Duplicate Project 1 (copy)')
+        self.assertEqual(
+            len(self.test_duplicate_activity_1.child_ids), 1
```
eilst, line 55: This is wrong: you should test new rec and whatever changes it makes....I think test_duplicate_activity_1. Why did you add this assert?

project_event/tests/test_project_event.py
eilst, line 58: What is the meaning of this assert

project_event/tests/test_project_event.py
eilst, line 50: You are very explicit renaming the methods:         self.write_main_task(vals) ->
        self.write_children(vals) -> self.write_activty_children(vals), thats good, you could put a more minighful and descriptive name to this test.

eilst (changes requested):

project_event/tests/test_project_event.py
jananjoy: @eilst  Ok changed

project_event/tests/test_project_event.py
jananjoy: @eilst ok

project_event/tests/test_project_event.py
jananjoy: @ ????

project_event/models/project_task.py
jananjoy: @eilst Good Catch

PR #397, opened 2019-03-21
[ADD]Test cases for error in duplicating Activities from Event model

jananjoy:


jananjoy:
This branch [Test cases for error in duplicating Activities from Event model] is integrated with the feature implementation pull request  @ [FIX]Error while replicating an Activities. So, I closed this pull request.

PR #400, opened 2019-03-25
[ADD] Timesheet task integration

rimbendhaou:
Timesheet task integration

### DIFF
--- a/project_event/models/hr_timesheet.py
+++ b/project_event/models/hr_timesheet.py
@@ -1,0 +1,32 @@
+# © 2018 Savoir-faire Linux
+# License LGPL-3.0 or later (http://www.gnu.org/licenses/lgpl).
+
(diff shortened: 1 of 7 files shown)


project_event/models/hr_timesheet.py
```diff
@@ -0,0 +1,29 @@
...
+from odoo import _, api, models
+from odoo.exceptions import ValidationError
+
+
+class AccountAnalyticLine(models.Model):
+    _inherit = 'account.analytic.line'
+
+    @api.model
+    def create(self, vals):
+        """Override to get the parent project of the task."""
```
eilst: Do we need this comment?  
Create and write should always be as clean as possible. so, maybe it's better that you create a new method: 
`def get_project_parent_of_task(self)....`

eilst (approved):

PR #401, opened 2019-03-26
[UPDATE] Refactoring field name  ''total_time' to 'actual_total_time' in  V.11

jananjoy:


project_event/models/project_task.py
```diff
@@ -265,15 +263,15 @@ def _compute_actual_total_time(self):
                         self.real_date_start))
                 hours = time_diff.hours
                 minutes = time_diff.minutes
-                self.total_time = "{0:0=2d}".format(
+                self.actual_total_time = "{0:0=2d}".format(
```
eilst, line 266: The report also uses this field, you should refactor there too

eilst:
Please also specify what is the new feature, even if it's really small

project_event/models/project_task.py
jananjoy: @eilst  Change in  report  project_task_activity_report file too

PR #402, opened 2019-03-26
[FIX]  Test case  for Actual Total time  computation 

jananjoy:


project_event/models/project_task.py
```diff
@@ -224,6 +224,7 @@ class Task(models.Model):
         string='Actual End Time',
     )
     actual_total_time = fields.Char(
+
```
eilst: Remove empty line

project_event/views/project_task_view.xml
```diff
@@ -1,15 +1,38 @@
 <?xml version="1.0" encoding="utf-8"?>
 <odoo>
+    <record id="project_event_task_tree_technical_sheet" model="ir.ui.view">
```
eilst: You should pull the latest changes, this was removed and you are adding it back. Please Pull the latest 11.0_develop and use it as base

project_event/models/project_task.py
```diff
@@ -224,6 +224,7 @@ class Task(models.Model):
         string='Actual End Time',
```
eilst: Please dont do "git add . " You should always chose only the files that you modify, not all.

project_event/models/project_task.py
jananjoy: @eilst Changed

project_event/views/project_task_view.xml
jananjoy: @eilst Ok

project_event/models/project_task.py
jananjoy: @eilst  Ok

eilst:
LVGTM

PR #416, opened 2019-04-01
[ADD] Added non_bookable_equipment_ids field to fieldNames in case of…

lgharib:
Added non_bookable_equipment_ids field to fieldNames in case of…

### DIFF
--- a/project_resource_calendar/models/calendar_event.py
+++ b/project_resource_calendar/models/calendar_event.py
@@ -27,7 +27,7 @@ class CalendarEvent(models.Model):
         string='Floor',
         related='room_id.floor',
     )
-    allow_double_book = fields.Boolean(
+    allow_room_double_booking = fields.Boolean(
         string='Allow double book',
         related='room_id.allow_double_book',
     )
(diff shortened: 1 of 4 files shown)


project_resource_calendar/models/calendar_event.py
```diff
@@ -116,6 +126,14 @@ class CalendarEvent(models.Model):
         'Current ID',
     )
 
+    @api.one
```
eilst, line 129: @api.depends('allow_double_book') or @api.depends('room_id'), right?.. so it is re computed when the fields that this one depends on one change

project_resource_calendar/models/calendar_event.py
```diff
@@ -27,11 +27,21 @@ class CalendarEvent(models.Model):
         string='Floor',
         related='room_id.floor',
     )
+    allow_double_book = fields.Boolean(
```
eilst: Not sure if this is the best name for the field. is it? Because we are checking for the double booking of the room. But if it is tightly related with the double book of the event it might be ok... so this is just a comment.

lgharib: Yes you're right we should call it room_double_bookable or allow_room_double_booking

project_resource_calendar/models/calendar_event.py
lgharib: Yes I added it, when we change the room_id the equipments changes also

eilst:
Properties were transferred to another PR

PR #418, opened 2019-04-01
[UPD] Weekly Report removed from print menu

eilst:
\


rimbendhaou (commented):
We still have Weekly Events Report in the action of an event form. It doesnt raise error but Check with functional if we need this action in an event form or not.

rimbendhaou (approved):

PR #419, opened 2019-04-01
[FIX] Change in the Information contents for Task in calender view 

jananjoy:


project_event/models/project_task.py
```diff
@@ -227,6 +227,14 @@ class Task(models.Model):
         string='Actual Total Time',
         compute='_compute_actual_total_time',
     )
+    name_code_pair = fields.Char(
```
rimbendhaou: We already have  a compute field : complete_name. This field concatenates the code/name

rimbendhaou (changes requested):

project_event/models/project_task.py
jananjoy: @rimbendhaou Changed

rimbendhaou (approved):

PR #421, opened 2019-04-02
[FIX] Traceback Real Date not present

eilst:


project_event/report/project_task_activity_report.xml
```diff
@@ -80,7 +80,12 @@
                 <td> <t t-esc="task['department']"/></td>
                 <td> <t t-esc="task['employee']"/></td>
                 <td> <t t-esc="datetime.datetime.strptime(task['expected_start'], '%Y-%m-%d %H:%M:%S').strftime('%H:%M')"/></td>
-                <td> <t t-esc="datetime.datetime.strptime(task['real_start'], '%Y-%m-%d %H:%M:%S').strftime('%H:%M')"/></td>
+                <t t-if="task['real_start']">
```
rimbendhaou, line 83: What about the other fields? Do they need check?

eilst: No because they are required fields

eilst: I will add tests for this possible arising issue

rimbendhaou (approved):

PR #424, opened 2019-04-10
[FIX] Prevent the removal of a clone calendar.event

jananjoy:


project_event/models/project_task.py
```diff
@@ -980,3 +979,14 @@ def get_confirmation_wizard(self, action):
             'target': 'new',
             'res_id': new_wizard.id,
         }
+
+    @api.multi
+    def unlink(self):
+        for rec in self:
+            if rec.task_state in \
```
rimbendhaou: better to check if it is not in draft

project_event/models/project_task.py
```diff
@@ -980,3 +979,14 @@ def get_confirmation_wizard(self, action):
...
+                    ['option', 'requested', 'read',
+                     'postponed', 'accepted', 'approved', 'done']:
+                value = self.env['calendar.event'].search([
```
rimbendhaou: check if we have a clone or not

project_resource_calendar/models/calendar_event.py
```diff
@@ -305,3 +305,15 @@ def validate_client_id_write(self, vals):
             if 'partner_ids' in vals and self.client_id.id not in vals['partner_ids'][0][2]:
                 vals['partner_ids'] = [
                     (6, 0, vals['partner_ids'][0][2] + [self.client_id.id])]
+
+    @api.multi
+    def unlink(self):
+        if self.event_task_id.task_state in \
```
rimbendhaou: why you check the state of the related project.task ?

rimbendhaou: if it is api.multi use the for rec in self:

project_event/models/project_task.py
rimbendhaou: what about canceled state?

project_resource_calendar/models/calendar_event.py
rimbendhaou: @jananjoy keep only this unlink, we dont need the other unlink for the moment. We need to prevent the delete of the clone only

rimbendhaou: and add the equivalent test

rimbendhaou (changes requested):

project_resource_calendar/models/calendar_event.py
```diff
@@ -305,3 +307,14 @@ def validate_client_id_write(self, vals):
...
+    @api.multi
+    def unlink(self):
+        for record in self:
+            if record.event_task_id:
+                    raise ValidationError(
+                                _(
+                                    self.get_error_type('TASK_CLONE_ERROR')
+                                )
+                            )
+        return super(CalendarEvent, self).unlink()
```
eilst, line 320: inside for: 
               return super(CalendarEvent, record).unlink()

project_resource_calendar/tests/test_calendar_event.py
```diff
@@ -181,3 +182,28 @@ def test_110_calendar_event_floor(self):
             self.post_room_id.floor,
             self.calendar_event1.room_floor
         )
+
+    def test_120_unlink(self):
```
eilst: provide a more descriptive test name

project_resource_calendar/tests/test_calendar_event.py
```diff
@@ -181,3 +182,36 @@ def test_110_calendar_event_floor(self):
...
+        self.room_1 = self.Rooms.create({
+            'name': 'Test Room 1',
+            'resource_type': 'room',
+            'allow_double_book': True,
+        })
+        self.partner_1 = self.Partners.create({
+            'name': 'Partner 1',
+            'email': 'test@test.com',
+        })
+        self.category_1 = self.Task_category.create({
```
rimbendhaou: there is no task_category in the common.py file or in the setup

project_resource_calendar/tests/test_calendar_event.py
```diff
@@ -181,3 +182,36 @@ def test_110_calendar_event_floor(self):
...
+            'responsible_id': self.responsible_1.id,
+            'partner_id': self.partner_1.id,
+            'category_id': self.category_1.id,
+            'room_id': self.room_1.id,
+            'resource_type': 'room',
+            'date_start': fields.Datetime.to_string(datetime.today()),
+            'date_end': fields.Datetime.to_string(datetime.today() +
+                                                  timedelta(hours=4)),
+        })
+        self.task_1.action_option()
```
rimbendhaou: wizard confirmation is missing

project_resource_calendar/tests/test_calendar_event.py
```diff
@@ -181,3 +182,36 @@ def test_110_calendar_event_floor(self):
...
+        calendar_event = self.task_1.info_calendar_event()
```
rimbendhaou: check that we have successfully an event clone

rimbendhaou (changes requested):

project_resource_calendar/tests/common.py
```diff
@@ -13,6 +13,7 @@ def setUp(self):
         self.Instruments = self.env['resource.calendar.instrument']
         self.WeeklyReportWizards = self.env['weekly.report.wizard']
         self.Partners = self.env['res.partner']
+        self.Tasks = self.env['project.task']
```
rimbendhaou: Why you add this if you will not use it in the project_resource_calendar tests

project_resource_calendar/models/calendar_event.py
```diff
@@ -206,6 +206,8 @@ def get_error_type(self, type_error):
             error_msg = _('this resource is not bookable')
         if type_error == 'ROOM_TYPE_ERROR':
             error_msg = _('this room is not bookable')
+        if type_error == 'TASK_CLONE_ERROR':
+            error_msg = _('Clone task cannot be deleted')
```
rimbendhaou, line 210: Add the translation

rimbendhaou (changes requested):

project_event/tests/common.py
```diff
@@ -178,3 +178,15 @@ def setUp(self):
             'room_id': self.room_1.id,
             'notes': 'Some Task Notes ...',
         })
+        self.room_4 = self.Rooms.create({
```
rimbendhaou: Is there any room 3? why directly room4?

rimbendhaou: If we already have a room and a partner and a responsible, we can use them

rimbendhaou:
@wbeverly waiting for the warning message (fFr/Eng)

rimbendhaou (approved):

PR #427, opened 2019-04-19
[Fix] Fix automatic creation of events from template

rimbendhaou:
Fix automatic creation of events from template


eilst:
Hey @rimbendhaou, to not have so many PRs here. When we know the PR will have the label Work in progress, better to not do the PR, we can commit and do all the work. and just do the PR when it's ready. When we change the label to WIP after some changes are requested it makes sense to have it here. if we want to reference... we can always use the hash of the commit

eilst (approved):

PR #430, opened 2019-04-23
[FIX] Refactoring hard-coded value into value Initialization 

jananjoy:


project_event/models/project_task.py
```diff
@@ -6,9 +6,26 @@
 from odoo import _, api, fields, models
 from odoo.exceptions import ValidationError
 
+ZERO = 0
+ONE = 1
+EMPTY = ''
+MINUTES = 60
```
rimbendhaou: if we will follow your logic, it should be MINUTES_IN_HOUR like hour_in_day

project_event/models/project_task.py
```diff
@@ -627,7 +644,7 @@ def _compute_order_task(self):
     def get_task_order(task_ds, activity_ds, format):
         time_diff = datetime.strptime(task_ds, format) \
                     - datetime.strptime(activity_ds, format)
-        return time_diff.days * 24 * 60 + time_diff.seconds / 60
+        return time_diff.days * HOUR_IN_DAY * MINUTES + time_diff.seconds / 60
```
rimbendhaou: what about the /60

rimbendhaou: in this case the second 60 should be SECONDS_IN_MINUTE

project_event/models/project_task.py
rimbendhaou: and the second one SECONDS_IN_MINUTE = 60

project_event/models/project_task.py
```diff
@@ -871,16 +890,13 @@ def child_reservation(child):
     def get_message_body(action):
         switcher = {
             'draft': ' ',
-            'option': _('The following is optional and \
-                        appears as crosshatched on your calendar'),
-            'requested': _('The following is requested'),
-            'accepted': _('The following is approved'),
+            'option': _(MSG_OPTION),
```
eilst, line 903: Did you check this works for translations? @jananjoy

project_event/models/project_task.py
```diff
@@ -6,9 +6,27 @@
 from odoo import _, api, fields, models
 from odoo.exceptions import ValidationError
 
+ZERO = 0
```
eilst: no use zero=0 could this be different? @jananjoy

project_event/models/project_task.py
```diff
@@ -6,9 +6,27 @@
...
+ONE = 1
```
eilst: same

project_event/models/project_task.py
```diff
@@ -6,9 +6,27 @@
...
+EMPTY = ''
```
eilst: same

project_event/models/project_task.py
```diff
@@ -6,9 +6,27 @@
...
+MINUTES_IN_HOUR = 60
```
eilst, line 9: this is good, it makes sense. it's more explicit

project_event/models/project_task.py
jananjoy: @rimbendhaou  it works

eilst (approved):

PR #432, opened 2019-04-26
[FIX] Changing the code structure and organization in project.task

jananjoy:


project_event/models/project_task.py
```diff
@@ -14,6 +14,7 @@ class Task(models.Model):
     _inherit = ['project.task']
     _rec_name = 'complete_name'
 
+    # Declaration of the Basic types fields
```
eilst: remove the comment, we understand this

project_event/models/project_task.py
```diff
@@ -72,37 +90,20 @@ class Task(models.Model):
...
+        string='Actual End Time',
     )
-    user_id = fields.Many2one(
-        'res.users',
-        string='Created by',
-        default=lambda self: self.env.uid,
-        index=True,
-        track_visibility='onchange',
+
+    # Declaration of the selection types fields
```
eilst: ok, good 👍  for the first review but you could have added in the PR description because in the code I think it is not necessary

project_event/models/project_task.py
```diff
@@ -189,68 +234,46 @@ class Task(models.Model):
...
-                self.spectators = valid_spectators
-            else:
-                raise ValidationError(
-                    _(
-                        'Spectators value should be in the range[' +
-                        str(MIN_SPECTATORS_VALUES_LIMIT) +
-                        '-' +
-                        str(MAX_SPECTATORS_VALUES_LIMIT) +
-                        ']'))
+    # Declaration of inherited  methods
```
eilst: No need for this kind of comments, good for the PR so we can see what was your intent, but you could better added in the PR description. In the code it is explicit enoght

project_event/models/project_task.py
```diff
@@ -14,6 +14,7 @@ class Task(models.Model):
     _inherit = ['project.task']
```
eilst: So you reordered the code? it's nice, it's more readable.

project_event/models/project_task.py
jananjoy: @eilst  ok

project_event/models/project_task.py
jananjoy: @eilst ok

project_event/models/project_task.py
jananjoy: @eilst

project_event/models/project_task.py
jananjoy: @eilst

eilst (approved):

PR #442, opened 2019-05-09
[ADD] Added wizard to calendar event overlap

lgharib:


project_event/wizard/double_task_validation_wiz.py
```diff
@@ -0,0 +1,38 @@
+# @ 2018 Savoir-failre Linux
+# License LGPL-3.0 or Later (http://www.gnu.org/licenses/lgpl).
+
+from odoo import api, fields, models
+
+
+class DoubleTaskValidationWiz(models.TransientModel):
+    """Wizard Reservation Confirmation"""
+    _name = 'double.task.validation.wiz'
```
rimbendhaou: the class name is ambiguous: It seems like if you will do a double check of the task

project_resource_calendar/models/calendar_event.py
```diff
@@ -27,11 +27,21 @@ class CalendarEvent(models.Model):
...
+        string='Allow double book',
+        related='room_id.allow_double_book',
+    )
     equipment_ids = fields.Many2many(
         string='Equipment',
         comodel_name='resource.calendar.instrument',
         ondelete='set null',
     )
+    double_bookable_equipment_ids = fields.Many2many(
+        string='Double bookable bookable equipment',
```
rimbendhaou:
```suggestion
        string='Double bookable equipments',
```

project_resource_calendar/models/calendar_event.py
```diff
@@ -352,3 +371,27 @@ def unlink(self):
                     )
                 )
         return super(CalendarEvent, self).unlink()
+
+    @api.multi
+    def get_confirmation_wizard(self, action):
+        self.ensure_one()
+        res = ''
+        if res != '':
```
rimbendhaou: res is always = ' '

project_resource_calendar/wizard/__init__.py
```diff
@@ -3,3 +3,4 @@
 
 from . import weekly_report_wizard
 from . import weekly_report
+from . import reservation_validation_wiz
```
rimbendhaou:
```suggestion
from . import reservation_validation_wiz

```

project_resource_calendar/wizard/reservation_validation_wiz.py
```diff
@@ -0,0 +1,38 @@
+# @ 2018 Savoir-failre Linux
+# License LGPL-3.0 or Later (http://www.gnu.org/licenses/lgpl).
+
+from odoo import api, fields, models
+
+
+class DoubleBookingValidationWiz(models.TransientModel):
```
rimbendhaou, line 7: this class looks like the DoubleTaskValidationWiz class except one field. Better to use the same class by adding the task_id field and checking if it is task or event

rimbendhaou (changes requested):
Check the comments

eilst:
@lgharib  Rebase to 11.0 or merge new changes, so you get the test coverage report

project_resource_calendar/wizard/reservation_validation_wiz.py
lgharib: Done

project_resource_calendar/models/calendar_event.py
lgharib: Removed the function because unused

project_resource_calendar/models/calendar_event.py
```diff
@@ -116,6 +126,15 @@ class CalendarEvent(models.Model):
         'Current ID',
     )
 
+    @api.one
+    @api.depends('room_id')
+    def _get_double_bookable_equipments(self):
+        double_bookable_equipment_ids = []
+        for equipment in self.equipment_ids:
```
eilst, line 133: you could replace this with a search

project_resource_calendar/wizard/reservation_validation_wiz.py
```diff
@@ -0,0 +1,62 @@
...
+        if self.event_id.id is not False:
+            self.event_id.write({'start_datetime': self.start_date,
+                                 'stop_datetime': self.end_date})
+        elif self.task_id.id is not False:
+            self.task_id.write({'date_start': self.start_date,
+                                'date_end': self.end_date})
+
+    @api.multi
+    def confirm_cancel(self):
+        if self.event_id.id is not False:
```
eilst:
```suggestion
        if self.event_id.id:
```

project_resource_calendar/wizard/reservation_validation_wiz.py
```diff
@@ -0,0 +1,62 @@
...
+            self.event_id.write({'start_datetime': self.r_start_date,
+                                 'stop_datetime': self.r_end_date})
+        elif self.task_id.id is not False:
```
eilst:
```suggestion
        elif self.task_id.id:
```

project_resource_calendar/wizard/reservation_validation_wiz.py
```diff
@@ -0,0 +1,62 @@
...
+        string='Rollback Starting Date'
+    )
+
+    r_end_date = fields.Datetime(
+        string='Rollback  Ending Date'
+    )
+
+    @api.multi
+    def confirm_update(self):
+        if self.event_id.id is not False:
```
eilst:
```suggestion
        if self.event_id.id:
```

project_resource_calendar/wizard/reservation_validation_wiz.py
```diff
@@ -0,0 +1,62 @@
...
+            self.event_id.write({'start_datetime': self.start_date,
+                                 'stop_datetime': self.end_date})
+        elif self.task_id.id is not False:
```
eilst:
```suggestion
        elif self.task_id.id:
```

project_resource_calendar/wizard/reservation_validation_wiz.py
eilst: These lines are very similar to the ones in the method above, you could extract a new method with dates as arguments and remove duplication

project_resource_calendar/static/src/js/update_event_wizard.js
```diff
@@ -0,0 +1,144 @@
...
+            if(self.modelName == 'calendar.event') {
+                var overlap = false;
+                var start_date = '';
+                var end_date = '';
+                var r_start_date = '';
+                var r_end_date = '';
+                var record_id = 0;
+                record.source.events.forEach(function(event) {
+                    if (event.id != record.id ){
+                        if(record._start._d < event._end._d && record._end._d > event._start._d){
```
eilst, line 38: could you merge some of this ifs?

project_resource_calendar/wizard/reservation_validation_wiz.py
lgharib: When self.task_id.id exist it's not true it returns an integer. So to test case when task_id was not filled we have to check if it's not False

rimbendhaou (approved):

PR #443, opened 2019-05-10
[FIX]Punctual Reservation Report: remove the header that appears at the top

jananjoy:


eilst (approved):

rimbendhaou (changes requested):
You pushed the modification of another task

rimbendhaou (approved):

PR #473, opened 2019-05-31
[FIX] fix message prompt

lgharib:
11 lmg fix message prompt

### DIFF
--- a/project_event/models/project_task.py
+++ b/project_event/models/project_task.py
@@ -755,6 +755,13 @@ class Task(models.Model):
                 overlaps_equipment.remove(overlap_equipment)
         if len(overlaps_equipment) > 0:
             booked_resources.append(self.equipment_id.name)
+
+        for attendee in self.get_partners():
+            hres = self.is_hr_resource_booked(attendee)
+            partner_attendee = self.env['res.partner'].browse(attendee)
+            if hres and partner_attendee:
+                res += partner_attendee.name + '<br>'
+
         return booked_resources
 
     @api.multi
--- a/project_event/models/project_task.py
+++ b/project_event/models/project_task.py
@@ -701,7 +701,7 @@ class Task(models.Model):
             res += self.room_id.name + '<br>' if (
                 self.room_id) else self.equipment_id.name + '<br>'
         for attendee in self.get_partners():
-            hres = self.is_hr_resource_booked(attendee)
+            hres = self.is_hr_resource_double_booked(attendee)
             partner_attendee = self.env['res.partner'].browse(attendee)
             if hres and partner_attendee:
                 res += partner_attendee.name + '<br>'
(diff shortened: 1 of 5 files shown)


project_event/models/project_task.py
```diff
@@ -1009,6 +1050,22 @@ def is_hr_resource_booked(self, attendee):
         ])
         return len(overlaps_partners) > 0
 
+    def is_hr_resource_double_booked(self, attendee, date_start, date_end):
```
rimbendhaou: For the method is_hr_resource_booked, i forget to check if it is not the same event_task_id, so when added we have to merge the 2 methods and keep only one

project_resource_calendar/models/calendar_event.py
```diff
@@ -371,3 +352,54 @@ def unlink(self):
...
         return super(CalendarEvent, self).unlink()
+
+    def is_hr_resource_double_booked(self, attendee,
+                                     date_start=None, date_end=None):
+        if not date_end and not date_start:
+            date_start = self.date_start
+            date_end = self.date_end
+
+        overlaps_partners = self.env['calendar.event'].search([
+            ('id', '!=', self.id),
```
rimbendhaou, line 363: we need to factor this redandant code, we can use a constatnt (OVERLAP_DOMAIN = [
            ('id', '!=', self.id),('start', '<', date_end),
            ('stop', '>', date_start),
            ('state', '!=', 'cancelled'),] and use it each time (we simply add the specific fields to check ( ('partner_ids', 'in', attendee.id))

rimbendhaou (changes requested):

project_event/models/project_task.py
```diff
@@ -727,6 +727,50 @@ def get_booked_resources(self):
                             '<br>'
         return res
 
+    def get_double_booked_resources(self, date_start=None, date_end=None):
```
eilst, line 730: hey, this methods are duplicated code. Just a comment, but you could reference the object and without creating it and make these verifications.

project_resource_calendar/models/calendar_event.py
```diff
@@ -371,3 +352,50 @@ def unlink(self):
...
+            ('partner_ids', 'in', attendee.id),
+            ('start', '<', date_end),
+            ('stop', '>', date_start),
+            ('state', '!=', 'cancelled'),
+        ])
+        return len(overlaps_partners) > 0
+
+    def get_double_booked_resources(self, date_start=None, date_end=None):
+        booked_resources = []
+        overlap_domain = [
```
rimbendhaou: overlap_domain should be after (if not date_end and not date_start:)

rimbendhaou (changes requested):

rimbendhaou (approved):

PR #490, opened 2019-06-26
[FIX] Fixed error on update list view activity double booking

lgharib:
Fixed error on update list view activity double booking


eilst (approved):
Code LGTM

PR #491, opened 2019-06-26
[FIX] Fixed recurrency overlap issue search by id

lgharib:
Fixed recurrency overlap issue search by id

### DIFF
--- a/project_resource_calendar/models/calendar_event.py
+++ b/project_resource_calendar/models/calendar_event.py
@@ -222,23 +222,23 @@ class CalendarEvent(models.Model):
                             if resource.id == record.room_id.id:
                                 message += self.fill_validation_message(
                                     resource.name,
-                                    event.start,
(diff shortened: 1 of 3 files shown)


project_resource_calendar/models/calendar_event.py
```diff
@@ -387,13 +387,18 @@ def is_hr_resource_double_booked(self, attendee,
...
         overlaps_partners = self.env['calendar.event'].search([
-            ('id', '!=', self.id),
             ('partner_ids', 'in', attendee),
             ('start', '<', date_end),
             ('stop', '>', date_start),
             ('state', '!=', 'cancelled'),
         ])
-        return len(overlaps_partners) > 0
+
+        overlaps_counter = len(overlaps_partners)
```
eilst: You could extract this method to avoid repetition

eilst (approved):

PR #495, opened 2019-07-02
[FIX] Fixed today date format

lgharib:
Fixed today date format

### DIFF
--- a/project_resource_calendar/wizard/weekly_report.py
+++ b/project_resource_calendar/wizard/weekly_report.py
@@ -15,8 +15,7 @@ class ReportWeekly(models.AbstractModel):
         room = self.env['resource.calendar.room'].browse(
             int(data['form']['room_id'])
         )
-        today = self.get_formatted_date(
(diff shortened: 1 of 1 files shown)


project_resource_calendar/wizard/weekly_report.py
```diff
@@ -75,11 +74,9 @@ def review_weekdays(events):
     def get_formatted_date(self, date):
         lang = self.env['res.users'].browse(self.env.uid).lang or 'en_US'
         tz = self.env['res.users'].browse(self.env.uid).tz or 'utc'
-        date_to_format = datetime.strptime(
```
eilst, line 78: mmm... I remember this method, I created last week. My fault I didn't add some tests, but I' m almost certain something will break with this change because since you change the obj type of the arg, we should also change other calls to this method.

eilst: It is ok, I have mistaken this method with the one in project_task

eilst (approved):

PR #499, opened 2019-07-04
[ADD] Added filter on equipement instrument technical sheet

lgharib:
Added filter on equipement instrument technical sheet

### DIFF
--- a/project_event/report/project_task_activity_report.py
+++ b/project_event/report/project_task_activity_report.py
@@ -53,6 +53,9 @@ class ReportWeekly(models.AbstractModel):
                     'department': task.department_id.name,
                     'expected_start': self.get_tz_format(task.date_start),
                     'employee': employee.name,
+                    'resource': task.equipment_id.name
+                    if task.equipment_id.category_type == 'service'
+                    else _('Not a service'),
                     'order': task.task_order,
                     'real_start':
                     self.get_tz_format(task.real_date_start) if (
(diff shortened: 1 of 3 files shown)


project_event/report/project_task_activity_report.py
```diff
@@ -78,17 +79,13 @@ def get_tasks_details(self, tasks):
                     'resource': resources_list,
                 })
             elif task.resource_type == 'equipment' and task.equipment_id:
```
eilst: You can merge the two if statements. And is task.resource_type == 'equipment' the same as task..equipment_id.category_type == 'equipment'?

project_event/report/project_task_activity_report.py
```diff
@@ -53,6 +53,7 @@ def get_task_values(self, tasks):
                     'department': task.department_id.name,
                     'expected_start': self.get_tz_format(task.date_start),
                     'employee': employee.name,
+                    'resource': task.equipment_id.name,
```
eilst: This could be None. So None.name would cause a problem

lgharib: Added a condition to display the equipment name otherwise a text message Not a service is displayed instead.

project_event/report/project_task_activity_report.py
lgharib: I combined all the if condition to remove nested code

eilst (approved):

PR #502, opened 2019-07-09
[ADD] Added timesheet to technical sheet and fixed raw data for Rooms

lgharib:
Added timesheet to technical sheet and fixed raw data for Rooms


eilst (approved):
Looks good 👍

PR #505, opened 2019-07-16
[ADD] Added tasks descriptions to bt and updated titles and remove ma…

lgharib:
Added tasks descriptions to bt and updated titles and remove main task


eilst (approved):
LGTM

PR #516, opened 2019-07-30
[UPD] calendar event unlink method location

eilst:


lgharib:
You should reconcider moving the event_taks_id as it is used in the calendar event clone  title to guive link for the user to the original task in the project event module

eilst:
Its done, I did the change before your comment, because the tests failed. now they are good with event_taks_id in project_resource_calendar module.

eilst:
Closed, fix is being treated elsewhere.

PR #517, opened 2019-07-30
[UPD] Serice as new field

eilst:


lgharib:
You should add the service_id field to the task model template and xml

eilst:
> You should add the service_id field to the task model template and xml


Done

lgharib (approved):

PR #520, opened 2019-08-06
[FIX] FIX multiple write to fix delete multiple recurrent event

rimbendhaou:
FIX multiple write to fix delete multiple recurrent event


eilst (approved):
LGTM, added multi to method validate_client_id_write

PR #522, opened 2019-08-07
[ADD] Added contraint on date_start and date_end

lgharib:
Added contraint on date_start and date_end


lgharib:
I should add local hour based on users preferences and not UTC time

eilst (changes requested):
Looks good

eilst (approved):

PR #537, opened 2019-08-29
[UPD] Removed rooms and miscellaneous from list of used equipment

lgharib:
Removed equipment and instrument from list of used equipments

### DIFF
--- a/project_event/report/project_task_activity_report.xml
+++ b/project_event/report/project_task_activity_report.xml
@@ -198,11 +198,20 @@
               <th>Resource</th>
             </tr>
             <t t-foreach="tasks_details" t-as="resource">
-              <tr>
-                <td> <t t-esc="resource['task']"/></td>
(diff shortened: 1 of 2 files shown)


project_event/report/project_task_activity_report.xml
```diff
@@ -198,11 +198,20 @@
               <th>Resource</th>
             </tr>
             <t t-foreach="tasks_details" t-as="resource">
-              <tr>
-                <td> <t t-esc="resource['task']"/></td>
-                <td> <t t-esc="resource['resource_type']"/></td>
-                <td> <t t-raw="resource['resource']"/></td>
-              </tr>
+              <t t-if="resource['resource_type']=='Instrument'">
```
eilst, line 201: It's good. resource['resource_type'] is taking the resource type title, this should be changed in the future. Now it works, but it would not work in another language than English or french

eilst (approved):

PR #544, opened 2019-09-12
[ADD] Limit write if room not in user rooms

eilst:


project_event/security/project_security.xml
```diff
@@ -130,14 +119,26 @@
...
-        <field name="groups" eval="[(4, ref('group_project_event_editor')),   (4, ref('group_project_event_manager'))]"/>
+        <field name="groups" eval="[(4, ref('group_project_event_manager'))]"/>
         <field eval="1" name="perm_unlink"/>
         <field eval="1" name="perm_write"/>
         <field eval="1" name="perm_read"/>
         <field eval="1" name="perm_create"/>
     </record>
+
+    <record id="calendar_event_access_project_editor_all" model="ir.rule">
+        <field name="name">Calendar Event all access to editor</field>
```
rimbendhaou: not all access, only read access in this case

rimbendhaou (changes requested):

project_event/security/project_security.xml
eilst: 1=1 i mean, access to all records

eilst: For this kind of comment please just comment, not request change; so if you are not available, we can still merge it  if someone else does the review

rimbendhaou (approved):

project_event/security/project_security.xml
rimbendhaou:
```suggestion
        <field name="name">All Calendar Event read access to editor</field>
```

PR #558, opened 2019-09-29
[ADD] Added move activity tasks module

lgharib:
Added move activity tasks module


project_move_activity_tasks/models/project_task.py
```diff
@@ -0,0 +1,29 @@
...
+        )
+        return {
+            'name': 'Move activity tasks',
+            'type': 'ir.actions.act_window',
+            'view_type': 'form',
+            'view_mode': 'form',
+            'res_model': 'move.activity.tasks.wiz',
+            'target': 'new',
+            'res_id': new_wizard.id,
+        }
```
eilst: pep8

project_move_activity_tasks/wizard/project_move_activity_tasks_wizard.py
```diff
@@ -0,0 +1,53 @@
...
+    moving_type = fields.Selection(string='Type',
+                                   selection=[
+                                       ('interval', 'Interval moving'),
+                                       ('date', 'Date moving')
+                                   ],
+                                   default="interval")
+
+    @api.multi
+    def confirm_moving(self):
+        print('Test')
```
eilst: Looks Good, I will give it a thumbs up. But, you could simply pass or log rather than printing, it would be better

eilst (approved):

project_move_activity_tasks/wizard/project_move_activity_tasks_wizard.xml
```diff
@@ -0,0 +1,48 @@
...
+                <group attrs="{'invisible': [('moving_type','=', 'date')]}">
+                    <field name="moving_direction" widget="radio"/>
+                </group>
+                <group align="right" attrs="{'invisible': [('moving_type','=', 'interval')]}">
+                    <field name="moving_date" string="Moving date"/>
+                </group>
+                <notebook>
+                        <page string="Tasks">
+                            <field name="child_ids" string="Task" context="{
+                            'form_view_ref':'project_event.project_event_task_form',
```
eilst: I think the problem is: wrong view, it should be  a tree view. not a project_event.project_event_task_form

project_move_activity_tasks/wizard/project_move_activity_tasks_wizard.py
```diff
@@ -0,0 +1,122 @@
...
+
+    def _compute_get_activity_childs(self):
+        tasks = []
+        for child in self.activity_id.child_ids:
+            if not child.is_main_task:
+                tasks.append(child.id)
+
+        self.child_ids = [(6, False, tasks)]
+
+    def _move_tasks_with_interval(self, direction, child):
```
eilst: It can also be an activity, so you could change the argument 'child' to 'task'

project_move_activity_tasks/wizard/project_move_activity_tasks_wizard.py
```diff
@@ -0,0 +1,122 @@
...
+                minutes=self.minutes)).strftime('%Y-%m-%d %H:%M:%S'),
+            'date_end': (datetime.datetime.strptime(child.date_end,
+                                                    '%Y-%m-%d %H:%M:%S') +
+                         direction * datetime.timedelta(
+                days=self.days,
+                hours=self.hours,
+                minutes=self.minutes)).strftime('%Y-%m-%d %H:%M:%S')
+        })
+
+    def _move_tasks_to_date(self, child):
```
eilst: It can also be an activity, so you could change the argument 'child' to 'task'

project_move_activity_tasks/wizard/project_move_activity_tasks_wizard.py
```diff
@@ -0,0 +1,122 @@
...
+        moving_date = datetime.datetime.strptime(self.moving_date,
+                                                 '%Y-%m-%d %H:%M:%S')
+        date_start = datetime.datetime.strptime(child.date_start,
+                                                '%Y-%m-%d %H:%M:%S')
+        date_end = datetime.datetime.strptime(child.date_end,
+                                              '%Y-%m-%d %H:%M:%S')
+        child.write({
```
eilst: So you moving the activity and its children, all to same date? The requirement wasn't that with the new date, a delta was calculated (old date of the activity - new date) and apply that delta to its children

lgharib: I should validate the requirements. I think that the first functionality of interval could be used for this purpose or the user can move just the Activity/Main task by checking the related boolean and move the rest of the tasks separately in a second operation.

PR #579, opened 2019-11-01
[FIX] Fixed no message sent for employe in main task

lgharib:
Fixed no message sent for employe in main task


eilst (approved):

eilst:
This could be improved, but OK

PR #587, opened 2019-11-26
11 lg access editor calendar event

lgharib:
Added raise error on editor calendar event


eilst:
I think you have two options:
-The better one is to do it using a record rule;
- Maybe by adding missing conditions, failing tests will pass

Pep8

gaultp (approved):

PR #598, opened 2019-12-09
[FIX] Fixed technical sheet issues

lgharib:
Fixed technical sheet issues, employee name should be visible only once in list of convocations and only the first task of the day.

### DIFF
--- a/project_event/report/project_task_activity_report.py
+++ b/project_event/report/project_task_activity_report.py
@@ -93,8 +93,8 @@ class ReportWeekly(models.AbstractModel):
         done = set()
         result = []
         for element in sorted_list:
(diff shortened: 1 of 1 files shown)


project_event/report/project_task_activity_report.py
```diff
@@ -93,8 +93,8 @@ def get_different_department_and_employee(sorted_list):
         done = set()
         result = []
         for element in sorted_list:
-            if (element['employee'], element['department']) not in done:
-                done.add((element['employee'], element['department']))
+            if element['employee'] not in done:
```
eilst, line 96: I thought the set (employee, department) was a specification. This spec changed multiple times.
So, even if an employee has two different departments we only display one?

PR #617, opened 2020-01-06
[UPD] Updated license from LGPL-3 to AGPL-3

lgharib:


eilst:
close because it;s already done

PR #618, opened 2020-01-06
[UPD] Updated license from LGPL-3 to AGPL-3

lgharib:


eilst:
https://github.com/savoirfairelinux/project-addons/pull/625

PR #644, opened 2020-05-14
[ADD] project task main task's description  field edit condition

eilst:
Fixes issue of editing of the main task by project.event user when task is in ['requested', 'accepted'] and the task is of type 'main_task' by always removing description value during the write

\Introduced following functional problem:
If user makes a change of the description, he will not have any feedback telling him that his change was not taken into account.


eilst:
Protect 11.0.0

